Issue #22, opened 2023-04-03
Create Snapshots during build


andreespirela:
# Chore

- Create snapshots
- Bisect code (necessary for snapshot)

## Describe the chore

We want to generate snapshots and attach it to the binary so that most of the heavy start-up work gets compiled.

Issue #26, opened 2023-04-05
Refactor User Runtime and Main runtime into a single runtime


andreespirela:
# Chore

Currently there's a different runtime for Users and for the main application by Supabase. 

## Describe the chore

We need to merge these two runtimes into a single runtime and control the behavior through ops and states

andreespirela:
Requires #32

PR #32, opened 2023-04-06
Initial start-up snapshots

andreespirela:
This PR solves for #22.


crates/sb_core/js/user_runtime_loader.js
```diff
@@ -0,0 +1,9 @@
+// This file is meant to only have `userRuntimeCleanUp`
+// The code should address any user specific runtime behavior
+// As well as deletions
+
+function loadUserRuntime() {
+    delete globalThis.EdgeRuntime;
```
andreespirela, line 6: @inian @laktek Everything that's inside this function will be called when the "User runtime" is created. For example, this cleans up the existence of `EdgeRuntime` in the user worker while using the same snapshot.

andreespirela: Next step here is unifying the initialization code. Will be done in #26

examples/stripe-web-hook/index.ts
```diff
@@ -12,7 +12,7 @@ const corsHeaders = {
   "Access-Control-Allow-Headers": "authorization, x-client-info, apikey",
 };
 
-const stripe = Stripe(Deno.env.get("STRIPE_API_KEY"), {
+const stripe = Stripe("STRIPE_API_KEY", {
```
laktek, line 15: I guess we would need to change this to get the environment variable?

laktek (approved):
LGTM!

examples/stripe-web-hook/index.ts
andreespirela: Right

PR #51, opened 2023-04-13
Fix enforcement of execution duration

mlafeldt:
I noticed that requests might hang forever despite the timeout feature being in place.

This problem (and the fix) can be reproduced with a main worker with a `workerTimeoutMs` of e.g. 3000 and the following user worker:

```ts
import { serve } from "https://deno.land/std@0.131.0/http/server.ts";

serve((_req) => {
  while (true) {}

  return new Response("Hello");
});
```

### DIFF
--- a/crates/base/src/js_worker.rs
+++ b/crates/base/src/js_worker.rs
@@ -387,7 +387,8 @@ impl UserWorker {
             let future = async move {
                 tokio::select! {
                     _ = tokio::time::sleep(Duration::from_millis(worker_timeout_ms)) => {
-                        debug!("max duration reached for the worker. terminating the worker. (duration {})", human_elapsed(worker_timeout_ms))
+                        error!("max duration reached for the worker. terminating the worker. (duration {})", human_elapsed(worker_timeout_ms));
+                        thread_safe_handle.terminate_execution();
                     }
                     Some(val) = memory_limit_rx.recv() => {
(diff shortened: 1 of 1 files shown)


laktek (approved):
Ah good catch! This must have accidentally removed during a refactor.

laktek:
The recent refactor makes it tricky to merge this change. I'll raise a separate PR with this patch.

laktek:
@mlafeldt Closing this in favor of #52 (I have made you a co-author in that PR)

rajivsinclair:
Is this max execution duration customizable from `config.toml`?

laktek:
@rajivsinclair Not from Supabase CLI (it defaults to the same limit as production). But you can configure the limits if you directly run the edge-runtime.

rajivsinclair:
Thanks for clarifying! I'm hoping to just self-host my prod instance of Supabase and I want long-running edge functions (>600s). Will it be possible to configure that if I adjust the edge runtime that Supabase CLI initiated?

laktek:
@rajivsinclair yep, you would need to modify the main worker script running in edge-runtime container to desired limits (https://github.com/supabase/edge-runtime/blob/main/examples/main/index.ts#L88-L101)

rajivsinclair:
Got it. Thanks for following up!

PR #52, opened 2023-04-17
fix: correctly enforce max duration

laktek:

This is based on #51. Had to make a couple of changes to work it correctly with current main.


andreespirela (approved):
LGTM

PR #56, opened 2023-04-18
fix: user worker crash on preflight request

### LINKED ISSUE #50: Preflight OPTIONS requests fail with 408 - Request timeout
# Bug report

## Describe the bug

When using `supabase functions serve` to serve all functions simultaneously, the pre-flight 'OPTIONS' request initiated by a browser agent fails.
The server returns Status Code `408 - Request timeout` with a message body of `{"msg":"Request could not be processed by the server because it timed out or an error was thrown."}`

**The bug is triggered by the fact that the OPTIONS body is empty**

## To Reproduce

1. Use code snippet at [https://supabase.com/docs/guides/functions/cors#recommended-setup](https://supabase.com/docs/guides/functions/cors#recommended-setup) to setup CORS headers in the `serve` function
2. Send `OPTIONS` request to edge function

## Expected behavior

The server should return `200` with the CORS headers set appropriately.

## Screenshots

Capture d’écran 2023-04-13 à 15 22 38
Capture d’écran 2023-04-13 à 15 23 28

## System information

- OS: macOS
- Browser (if applies) : chrome. Also tested outside browser (Postman)
- Version of supabase-js: 2.20.0
- Version of Node.js: not relevant

## Additional context

1. If serving the function individually via `supabase functions serve <function-name>`, the bug does not happen. It only happens when serving all functions simultaneously via `supabase functions serve`

2. If calling the function with the `POST` method directly, the bug does not happen. 

3. Inserting `console.log()` statements everywhere in the edge function code shows that the script reaches the point where it has to send the response to the OPTIONS request, then crashes silently.

4. If you put some dummy data into the body of the `OPTIONS` request, the bug does not happen:
Capture d’écran 2023-04-13 à 15 24 00

*Unfortunately a browser-generated pre-flight request always has an empty body*

To me it looks like when I call `return new Response('ok', { headers: corsHeaders })` inside the edge function, this triggers the bug because probably the server-side is still waiting for the client to send its body and it does not want to reply with the Response before it can close the incoming connection - but I may be mistaken on my interpretation.

**In conclusion, it looks like there may be some code inside edge-runtime that tries to enforce a non-empty body on an `OPTIONS` request ?**

soedirgo:
Fixes #50

P.S. I can't get the test to work - can't run the tests locally either since fetch calls just fail, so might need some help there.

### DIFF
--- a/crates/sb_workers/lib.rs
+++ b/crates/sb_workers/lib.rs
@@ -389,10 +389,7 @@ impl Stream for MpscByteStream {
         let val = std::task::ready!(self.receiver.poll_recv(cx));
         match val {
             None if self.shutdown => Poll::Ready(None),
-            None => Poll::Ready(Some(Err(std::io::Error::new(
-                std::io::ErrorKind::UnexpectedEof,
-                "channel closed",
-            )))),
+            None => Poll::Ready(None),
             Some(None) => {
                 self.shutdown = true;
                 Poll::Ready(None)
--- a/test/spec/hello.spec.ts
+++ b/test/spec/hello.spec.ts
@@ -246,4 +246,17 @@ describe('basic tests (hello function)', () => {
     log(`assert ${data} is equal to 'Hello World'`)
     expect(data).toEqual('Hello World')
   })
+
+  // regression: https://github.com/supabase/edge-runtime/issues/50
+  test('invoke preflight request', async () => {
+    log('perform a preflight request')
+    const res = await getCustomFetch(`http://localhost:${relay.container.getMappedPort(port)}/hello`, {
+      method: 'OPTIONS',
+      headers: {
+        Authorization: `Bearer ${apiKey}`
+      }
+    })('')
+
+    expect(await res.text()).toMatchInlineSnapshot()
+  })
 })


laktek:
@soedirgo for the test, I think you still need to handle the `OPTIONS` request inside the `hello` function and return a response (located at test/functions/hello/index.ts)

soedirgo:
As per https://github.com/supabase/edge-runtime/issues/50#issuecomment-1513210650, I'm not sure if this fix is enough - we need a reliable way to check if a request has a body.

> What happens if you handle it in main function?

Will test that in a bit.

vejja:
> As per [#50 (comment)](https://github.com/supabase/edge-runtime/issues/50#issuecomment-1513210650), I'm not sure if this fix is enough - we need a reliable way to check if a request has a body.

I have absolutely no knowledge or understanding of the Deno runtime, but I am wondering why we are trying to detect in the first place whether or not the request has a body.
Wouldn't it be safer to forward the body 'as is', and leave it up to the ultimate handler function to decide what to do with it ?

soedirgo:
I'm also not too familiar with the runtime - detecting whether or not the request has a body is likely a red herring, the body probably isn't being streamed correctly from the JS code into the Rust code.

> What happens if you handle it in main function?

Still having the [issue](https://github.com/supabase/edge-runtime/actions/runs/4739247264/jobs/8413898015#step:8:335).

laktek:
Closing this in favor of #75

PR #57, opened 2023-04-19
Add formatting / Clippy

laktek:

This PR adds two scripts:

`scripts/fmt.sh` - for format code with rustfmt
`script/clippy.sh` - for running linting and code improvements using Clippy

This is also integrated to the CI so it runs automatically for new PRs.

Notes:

* I removed the tests from `module_fetcher` - as it's directly copied from deno without modification, and we don't run tests on it.
* Also, fixed some code based on Clippy suggestions


andreespirela (approved):
lgtm

PR #63, opened 2023-04-24
Refactor connection handling

laktek:

* Refactor handling of incoming requests to improve performance.
* When a worker is a shutdown, the pool gets notified so it knows not to route future requests to that worker.
* Updated the`examples/main` to reuse existing workers.

### DIFF
--- a/crates/base/src/edge_runtime.rs
+++ b/crates/base/src/edge_runtime.rs
@@ -189,15 +189,10 @@ impl EdgeRuntime {
 
     pub async fn run(
         mut self,
-        stream: UnixStream,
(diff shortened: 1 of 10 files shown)


andreespirela:
I feel like after this PR we should re-evaluate the way we're using `channels` and all the `.clone()` in the codebase. It can probably use some refactoring, and the more we can avoid `.clone()` the better for memory usage and performance growth

laktek:
Most of the time, `clone()` is used when we need to share resources safely with an async clousres. For channels, clones should be relatively cheap - under the hood they  are just async resource counters.

Also, Clippy would pick up unnecessary clones.

andreespirela (approved):
LGTM

Issue #60, opened 2023-04-21
Latest oak server causing edge runtime to throw an error


silentworks:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When using the latest (v12.2.0 at the time of writing) oak server in the edge runtime it throw an error about `worker thread panicked TypeError: Cannot read properties of undefined (reading 'querySync'). However if you should explicitly state the version of oak and use the previous version v12.1.0 this error goes away and the endpoints work as intended.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Deploy the [self-hosted-edge-functions-demo](https://github.com/supabase/self-hosted-edge-functions-demo) to fly.io or any hosting provider
2. Make a request to any of the oak server's endpoints
3. Check the console where you deployed the project to see the error

## Expected behavior

No errors, it should just work.

## Screenshots

![Screenshot 2023-04-21 004010](https://user-images.githubusercontent.com/79497/233516653-df68df7b-e455-45c7-b8ff-0bd77f52dff3.png)

## System information

- OS: Windows 11
- Browser (if applies) [e.g. chrome, safari]
- Version of edge-runtime: v1.2.14

laktek:
Looks like we may need to expose `Deno.permissions` global object. `querySync` is a property of it.

andreespirela:
<img width="245" alt="image" src="https://user-images.githubusercontent.com/60560109/234111754-daa4ea99-4bec-437f-8c7d-78f4230b00ba.png">
<img width="1443" alt="image" src="https://user-images.githubusercontent.com/60560109/234111795-1f4a3fff-4fa9-4c62-bdc2-d6347ba85f84.png">

/greet
<img width="1320" alt="image" src="https://user-images.githubusercontent.com/60560109/234111998-d2f5cd49-afe5-42bf-b50e-2b73d94687c1.png">

andreespirela:
Seems as mentioned above, it only happens with v12.2.0

andreespirela:
@laktek Fixed this bug, will open a PR. Needs some discussion & trying to write some tests, lil tricky.

andreespirela:
Fixed in #64 .

PR #64, opened 2023-04-24
Fix oak #60

andreespirela:

This PR introruces:
- Fixing the oak issue #60 
- Catching the error inside the EdgeResult 
- Tests
- Permissions exposed

### DIFF
--- a/crates/base/src/edge_runtime.rs
+++ b/crates/base/src/edge_runtime.rs
@@ -572,8 +572,8 @@ mod test {
     #[tokio::test]
     async fn test_oak_60_issue() {
         let user_rt = create_basic_user_runtime("./test_cases/oak", 5, 1000);
-        let (stream, shutdown, _receiver) = create_user_rt_params_to_run();
-        let data = user_rt.run(stream, shutdown).await.unwrap();
-        // assert_eq!(data, EdgeCallResult::HeapLimitReached);
+        let unix_stream_rx = create_user_rt_params_to_run();
+        let data = user_rt.run(unix_stream_rx).await.unwrap();
+        //assert_eq!(data, EdgeCallResult::HeapLimitReached);
     }
 }
--- a/crates/base/src/edge_runtime.rs
+++ b/crates/base/src/edge_runtime.rs
@@ -12,12 +12,12 @@ use import_map::{parse_from_json, ImportMap, ImportMapDiagnostic};
 use log::{debug, error, warn};
 use serde::de::DeserializeOwned;
 use std::collections::HashMap;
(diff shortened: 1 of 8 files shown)


crates/sb_core/js/permissions.js
```diff
@@ -0,0 +1,279 @@
...
+    "ffi",
+    "hrtime",
+];
+
+/**
+ * @param {Deno.PermissionDescriptor} desc
+ * @returns {Deno.PermissionState}
+ */
+function opQuery(desc) {
+    return "granted";
```
andreespirela, line 49: @laktek in Deno, this line calls an op part of the `runtime`. since it was runtime specific, the quickest work-around is returning what's considered "granted". So any permission is "granted" from the edge runtime perspective. Do we want to move forward with this approach or is there something to consider?

crates/base/src/edge_runtime.rs
```diff
@@ -542,4 +568,12 @@ mod test {
         let data = user_rt.run(unix_stream_rx).await.unwrap();
         assert_eq!(data, EdgeCallResult::HeapLimitReached);
     }
+
+    #[tokio::test]
+    async fn test_oak_60_issue() {
```
andreespirela, line 573: Need to fix this. 
Test is throwing
```worker thread panicked Uncaught null

called `Result::unwrap()` on an `Err` value: Uncaught null```

which makes no sense because it works with the server, and tests basically replicate the same behavior as a server request. Not sure what's going on. I'll take a look, otherwise i'll comment it

File: crates/base/src/edge_runtime.rs [hunk 3]
```diff
@@ -530,16 +558,51 @@ mod test {
...
         let user_rt = create_basic_user_runtime("./test_cases/heap_limit", 5, 1000);
-        let unix_stream_rx = create_user_rt_params_to_run();
+        let (tx, unix_stream_rx) = create_user_rt_params_to_run();
         let data = user_rt.run(unix_stream_rx).await.unwrap();
         assert_eq!(data, EdgeCallResult::HeapLimitReached);
     }
+
+    #[tokio::test]
+    async fn test_oak_60_issue() {
+        let main_rt_simulation = async {
```
andreespirela, line 576: @laktek Example of an integration test here

andreespirela:
Fixes #60

laktek (approved):

crates/sb_core/js/permissions.js
laktek: yeah, good for now. Let's revisit permissions handling in a future task.

PR #69, opened 2023-04-26
Fixing body streamers

### LINKED ISSUE #50: Preflight OPTIONS requests fail with 408 - Request timeout

andreespirela:

Fixes #50 .

This fix corrects the way we identify when a request has a body.

### DIFF
--- a/crates/sb_workers/user_workers.js
+++ b/crates/sb_workers/user_workers.js
@@ -18,16 +18,6 @@ const ops = core.ops;
 //     envVars?: Array<any>
 // }
 
-const createStream = (array) => {
-    return new ReadableStream({
-        start: (controller) => {
-            controller.enqueue(array);
-            controller.close();
-        }
-    });
-};
-
-
 class UserWorker {
     constructor(key) {
         this.key = key;
--- a/crates/sb_workers/user_workers.js
+++ b/crates/sb_workers/user_workers.js
@@ -27,7 +27,7 @@ class UserWorker {
         const { method, url, headers, body } = req;
 
         const headersArray = Array.from(headers.entries());
-        const hasBody = headersArray.some(([headerName, headerValue]) => headerName.toLowerCase() === 'content-length' && Number(headerValue) > 0) || body == null;
+        const hasBody = headersArray.some(([headerName, headerValue]) => headerName.toLowerCase() === 'content-length' && Number(headerValue) > 0) && body !== null;
 
         const userWorkerReq = {
             method,
--- a/crates/sb_workers/user_workers.js
+++ b/crates/sb_workers/user_workers.js
@@ -36,11 +36,8 @@ class UserWorker {
     async fetch(req) {
         const { method, url, headers, body } = req;
 
-        const bodyData = (await new Response(body).arrayBuffer());
-        const bodyLength = bodyData.byteLength;
-
         const headersArray = Array.from(headers.entries());
-        const hasBody = bodyLength > 0;
+        const hasBody = headersArray.some(([headerName, headerValue]) => headerName.toLowerCase() === 'content-length' && Number(headerValue) > 0);
 
         const userWorkerReq = {
             method,
@@ -54,7 +51,7 @@ class UserWorker {
         // stream the request body
         if (hasBody) {
             let writableStream = writableStreamForRid(requestBodyRid);
-            await (createStream(new Uint8Array(bodyData))).pipeTo(writableStream);
+            await body.pipeTo(writableStream);
         }
 
         const res = await core.opAsync("op_user_worker_fetch_send", this.key, requestRid);
--- a/crates/sb_workers/user_workers.js
+++ b/crates/sb_workers/user_workers.js
@@ -18,16 +18,29 @@ const ops = core.ops;
 //     envVars?: Array<any>
 // }
 
+const createStream = (array) => {
+    return new ReadableStream({
+        start: (controller) => {
+            controller.enqueue(array);
+            controller.close();
+        }
+    });
+};
+
+
 class UserWorker {
     constructor(key) {
         this.key = key;
     }
 
     async fetch(req) {
-        const { method, url, headers, body, bodyUsed } = req;
+        const { method, url, headers, body } = req;
+
+        const bodyData = (await new Response(body).arrayBuffer());
+        const bodyLength = bodyData.byteLength;
 
         const headersArray = Array.from(headers.entries());
-        const hasBody = !(method === "GET" || method === "HEAD" || bodyUsed);
+        const hasBody = bodyLength > 0;
 
         const userWorkerReq = {
             method,
@@ -41,7 +54,7 @@ class UserWorker {
         // stream the request body
         if (hasBody) {
             let writableStream = writableStreamForRid(requestBodyRid);
-            body.pipeTo(writableStream);
+            await (createStream(new Uint8Array(bodyData))).pipeTo(writableStream);
         }
 
         const res = await core.opAsync("op_user_worker_fetch_send", this.key, requestRid);
--- a/crates/sb_workers/user_workers.js
+++ b/crates/sb_workers/user_workers.js
@@ -27,7 +27,7 @@ class UserWorker {
         const { method, url, headers, body } = req;
 
         const headersArray = Array.from(headers.entries());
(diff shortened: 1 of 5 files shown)


crates/sb_workers/user_workers.js
```diff
@@ -18,16 +18,29 @@ const ops = core.ops;
...
 class UserWorker {
     constructor(key) {
         this.key = key;
     }
 
     async fetch(req) {
-        const { method, url, headers, body, bodyUsed } = req;
+        const { method, url, headers, body } = req;
+
+        const bodyData = (await new Response(body).arrayBuffer());
```
andreespirela, line 39: This fix is a little odd. cc @laktek Let's get some eyes here.

vejja:
@andreespirela @laktek : sorry to intervene in a field where I certainly don't get all the complexities, but I do feel that `hasBody` is a bit hacky here. Especially relying on parsing the headers and looking up for `content-length` which is not guaranteed to be correct.

Why not rely on the standard [`fetch`](https://fetch.spec.whatwg.org/#ref-for-dom-body-bodyused) API spec? 
It says that:
> The body getter steps are to return null if [this](https://webidl.spec.whatwg.org/#this)’s [body](https://fetch.spec.whatwg.org/#concept-body-body) is null; otherwise [this](https://webidl.spec.whatwg.org/#this)’s [body](https://fetch.spec.whatwg.org/#concept-body-body)’s [stream](https://fetch.spec.whatwg.org/#concept-body-stream).
>
> The bodyUsed getter steps are to return true if [this](https://webidl.spec.whatwg.org/#this)’s [body](https://fetch.spec.whatwg.org/#concept-body-body) is non-null and [this](https://webidl.spec.whatwg.org/#this)’s [body](https://fetch.spec.whatwg.org/#concept-body-body)’s [stream](https://fetch.spec.whatwg.org/#concept-body-stream) is [disturbed](https://streams.spec.whatwg.org/#is-readable-stream-disturbed); otherwise false.

In summary, `hasBody` should be equal to

- `body` is not null
- AND : `bodyUsed` is false

Am I mistaken here?

crates/sb_workers/user_workers.js
```diff
@@ -36,11 +36,8 @@ class UserWorker {
     async fetch(req) {
         const { method, url, headers, body } = req;
```
vejja, line 27: const { method, url, headers, body, bodyUsed } = req;

crates/sb_workers/user_workers.js
```diff
@@ -36,11 +36,8 @@ class UserWorker {
...
 
-        const bodyData = (await new Response(body).arrayBuffer());
-        const bodyLength = bodyData.byteLength;
-
         const headersArray = Array.from(headers.entries());
-        const hasBody = bodyLength > 0;
+        const hasBody = headersArray.some(([headerName, headerValue]) => headerName.toLowerCase() === 'content-length' && Number(headerValue) > 0);
```
vejja, line 30: const hasBody = (body !== null) && (!bodyUsed);

vejja (changes requested):

crates/sb_workers/user_workers.js
andreespirela: I will add the `body !== null`. `!(bodyUsed)` is unnecessary, but checking if body is null is okay. Note that for the time being, `body` will always have a value. It's supposed to be null when it's empty, but the tech behind this is not respecting the spec. It's an upstream

vejja: The reason for checking !bodyUsed is that if the body stream has already been read, you are piping a closed stream to the Rust worker, which I would imagine could trigger all sorts of errors

vejja: > `body` will always have a value. It's supposed to be null when it's empty, but the tech behind this is not respecting the spec. It's an upstream

Where does the request come from ? Isn't it a JS object coming from Deno's std/http/server ?

vejja (changes requested):

crates/sb_workers/user_workers.js
andreespirela: At the time of using `bodyUsed` it's never been read. In fact we need to figure out whether `hasBody` is true or false before we even pipe it. If we for some reason consume the body and then pipe it, tests and everything will fail. Which is why it's designed -> Figure out whether it has a body _without consuming it_ -> Pipe it.

crates/sb_workers/user_workers.js
```diff
@@ -27,7 +27,7 @@ class UserWorker {
         const { method, url, headers, body } = req;
 
         const headersArray = Array.from(headers.entries());
-        const hasBody = headersArray.some(([headerName, headerValue]) => headerName.toLowerCase() === 'content-length' && Number(headerValue) > 0);
+        const hasBody = headersArray.some(([headerName, headerValue]) => headerName.toLowerCase() === 'content-length' && Number(headerValue) > 0) || body == null;
```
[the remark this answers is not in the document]
andreespirela: Yes, pushing change

vejja:
An alternative way of doing it
- if you can't rely on `body !== null` (because it always has a value)
- and if you are 100% sure that the stream has never been read at this point

would be to read the body stream at this stage, and then based on whether or not the resulting bytes array is empty, then create a brand new stream that you write down to the worker. 

It's a bit of a shame to force-read a stream to then re-write it, instead of elegantly piping it. But it might be more reliable than relying on `content-length` which the HTTP spec does not make mandatory - and even forbids when `transfer-encoding` is set.

What do you think?

vejja:
> An alternative way of doing it
> 
> * if you can't rely on `body !== null` (because it always has a value)
> * and if you are 100% sure that the stream has never been read at this point
> 
> would be to read the body stream at this stage, and then based on whether or not the resulting bytes array is empty, then create a brand new stream that you write down to the worker.
> 
> It's a bit of a shame to force-read a stream to then re-write it, instead of elegantly piping it. But it might be more reliable than relying on `content-length` which the HTTP spec does not make mandatory - and even forbids when `transfer-encoding` is set.
> 
> What do you think?

Reflecting here on my own suggestions - I think this is silly because it defeats the whole purpose of (1) streams and (2) HTTP chunked transfers... So I'm downvoting myself.

However I gave a quick look at Rust's `hyper` and it looks like if you call the `data()` function on a `Body`, it will return `None` if the body is empty. That might be the safest way to perform the test, however it must be done at the Rust level, not the JS one...

laktek:
Thanks @andreespirela! I opened a PR with a slightly different implementation (#75), which handles the case where request body can be a chunked stream. I will close this in favor of it.

PR #70, opened 2023-04-27
fix: enable performance API

laktek:

This PR enables the use of Performance API (https://developer.mozilla.org/en-US/docs/Web/API/Performance). We disable the use high-res time to protect against timing attacks and fingerprinting (https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#security_requirements).


andreespirela (approved):
LGTM

PR #74, opened 2023-04-29
Json modules coming to Supabase edge runtime

### LINKED ISSUE #30: Cannot import JSON dependencies
# Bug report

Try to import the following dependency within a function.

```
import cl100k_base from 'https://esm.sh/@dqbd/tiktoken@1.0.2/encoders/cl100k_base.json' assert { type: 'json' }

console.log(cl100k_base)
```

It throws the following error when the function is run:

```
worker thread panicked Expected ';', '}' or <eof> at https://esm.sh/@dqbd/tiktoken@1.0.2/encoders/cl100k_base.json:1:11
```

andreespirela:

This PR fixes #30, enabling json modules to be imported.


crates/base/src/js_worker/module_loader.rs
```diff
@@ -143,15 +143,30 @@ impl ModuleLoader for DefaultModuleLoader {
...
-                &code,
-                &emit_options,
-                emit_config_hash,
-            )?;
+            let code = match fetched_file.media_type {
+                MediaType::JavaScript
+                | MediaType::Unknown
+                | MediaType::Cjs
+                | MediaType::Mjs
+                | MediaType::Json => code.into(),
```
andreespirela, line 151: This should also remove the need to always `emit_parsed_source`. But abstract it from doing so in files that aren't meant to be processed/emitted

laktek (approved):

Issue #54, opened 2023-04-17
Importing axios throws error


sweatybridge:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Unable to import axios from edge runtime.

## To Reproduce

- create and serve `supabase/functions/hello/index.ts`
```typescript
import { serve } from "https://deno.land/std@0.168.0/http/server.ts";
import axios from "https://esm.sh/axios@1.3.5";

serve(async (req) => {
  console.log("Hello from Functions!");
  const { data } = await axios.get("https://supabase.com");
  return new Response(JSON.stringify(data), {
    headers: { "Content-Type": "application/json" },
  });
});
```

- request error `curl http://localhost:54321/functions/v1/hello -X POST`
```bash
$ supabase functions serve --no-verify-jwt
Setting up Edge Functions runtime...
Serving functions on http://localhost:54321/functions/v1/<function-name>
serving the request with /home/deno/functions/hello
worker thread panicked TypeError: _.accessor is not a function
    at https://esm.sh/v115/axios@1.3.5/esnext/axios.mjs:679:3
```

## Expected behavior

No errors if the same function is served with deno directly, ie.
```bash
deno run --allow-net supabase/functions/hello/index.ts
```

## System information

- OS: [e.g. macOS, Windows]
- Browser (if applies) [e.g. chrome, safari]
- Version of supabase-js: [e.g. 6.0.2]
- Version of Node.js: [e.g. 10.10.0]

## Additional context

related https://github.com/hwchase17/langchainjs/issues/633

andreespirela:
Investigating

andreespirela:
```
var _ = class {
    constructor(t) {
        t && this.set(t)
    }
    set(t, r, n) {
        let o = this;

        function s(c, d, u) {
            let l = I(d);
            if (!l) throw new Error("header name must be a non-empty string");
            let p = a.findKey(o, l);
            (!p || o[p] === void 0 || u === !0 || u === void 0 && o[p] !== !1) && (o[p || d] = Y(c))
        }
        let i = (c, d) => a.forEach(c, (u, l) => s(u, l, d));
[...]

PR #77, opened 2023-05-01
chore: added Axios example

laktek:

Adding an example for using Axios in edge functions. I couldn't reproduce #54 with the example code.


andreespirela (approved):
Lgtm

PR #84, opened 2023-05-09
Fix body null for null status

andreespirela:

Fixes https://github.com/supabase/cli/issues/1071

### DIFF
--- a/crates/sb_workers/user_workers.js
+++ b/crates/sb_workers/user_workers.js
@@ -49,7 +49,7 @@ class UserWorker {
 
         const res = await core.opAsync("op_user_worker_fetch_send", this.key, requestRid);
         const bodyStream = readableStreamForRid(res.bodyRid);
-        return new Response(bodyStream, {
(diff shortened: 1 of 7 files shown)


crates/sb_workers/user_workers.js
```diff
@@ -49,7 +49,7 @@ class UserWorker {
 
         const res = await core.opAsync("op_user_worker_fetch_send", this.key, requestRid);
         const bodyStream = readableStreamForRid(res.bodyRid);
-        return new Response(bodyStream, {
+        return new Response(!hasBody ? undefined : bodyStream, {
```
laktek, line 52: this `hasBody` check is for the request body right? I don't think we can use that here.

Would be helpful if you can add a test case where request has a body (eg: POST) and response doesn't have a body (204 no content)

andreespirela: Oh you're right. I did write a test with `POST` and succeeded. Although, I'll look into better ways to do this. Pushing test rn

laktek (approved):

PR #87, opened 2023-05-10
Deno fs main rt only

andreespirela:

This chore adds Deno.[FS-APIS] to the Main Runtime only. A little bit of the magic of `loadUserRuntime` :)


crates/sb_core/js/user_runtime_loader.js
```diff
@@ -2,7 +2,20 @@
...
+}
+
 function loadUserRuntime() {
+    deleteDenoApis(['chdir', 'chmod', 'chmodSync', 'chown', 'chownSync', 'copyFile', 'copyFileSync', 'create', 'createSync', 'cwd', 'fdatasync',
+        'fdatasyncSync', 'File', 'flock', 'flockSync', 'FsFile', 'fstat', 'fstatSync', 'fsync', 'fsyncSync', 'ftruncate', 'ftruncateSync',
+        'funlock', 'funlockSync', 'futime', 'futimeSync', 'link', 'linkSync', 'lstat', 'lstatSync', 'makeTempDir', 'makeTempDirSync',
+        'makeTempFile', 'makeTempFileSync', 'mkdir', 'mkdirSync', 'open', 'openSync', 'readDir', 'readDirSync', 'readFile',
+        'readFileSync', 'readLink', 'readLinkSync', 'readTextFile', 'readTextFileSync', 'realPath', 'realPathSync', 'remove',
+        'removeSync', 'rename', 'renameSync', 'seek', 'seekSync', 'stat', 'statSync', 'symlink', 'symlinkSync', 'truncate',
+        'truncateSync', 'umask', 'utime', 'utimeSync', 'writeFile', 'writeFileSync', 'writeTextFile', 'writeTextFileSync']);
```
laktek, line 18: Is it possible the not hardcode this? Can we import fsVars from `denoOverrides.js` and use its keys?

laktek:
Good to add 2 integration tests - one ensuring main function can read a file and another for user function failing to read a file.

andreespirela:
> Good to add 2 integration tests - one ensuring main function can read a file and another for user function failing to read a file.

We've got two tests doing this at the Runtime creation level which really covers it. Do we also want to add actual integration?

crates/sb_core/js/user_runtime_loader.js
andreespirela: That should be fine

laktek:
Runtime tests should be fine.

laktek (approved):

PR #93, opened 2023-05-15
body stream fixes

### LINKED ISSUE #91: ReadableStream as Response.body causes function to stall
# Bug report

## Describe the bug

With Supabase CLI 1.57.4, the edge-runtime version 1.2.19 was rolled out.
This includes a fix for sending back `null` values in the HTTP 204 status code, which rquired no body to be sent. (See issue https://github.com/supabase/cli/issues/1071 and its fix)

I suspect that with this change, the support for `ReadableStream`s has been broken, as they don't contain a `.size` and thus might be treated incorrectly by the fix.

## To Reproduce

1. Go to https://github.com/zcei/supabase-response-stream-bug
2. Clone repository & setup supabase
3. Serve functions
4. Issue the two requests from the bottom of the README

## Expected behavior

Both requests should work: one serves back a 204 with no content and the other a 200 with the text content of the stream.

## ~Screenshots~

~If applicable, add screenshots to help explain your problem.~

## System information

- OS: macOS
- ~Browser (if applies) [e.g. chrome, safari]~ reproducible with Browsers as with cURL
- Version of supabase-js: 1.58.1
- Version of Node.js: 18.14.1

## Additional context

I suspect that the line to fix it ([see here](https://github.com/supabase/edge-runtime/pull/84/files#diff-3988aa73db0fdfca8e726d9dc95e7ed1f098d4805a0f3f1160d9b2464be25779R52)) does not account for all cases. e.g. I don't think a `ReadableStream` can have a `.size` property when it's not clear yet whether someone will write to that stream.

I was trying to hop through Deno with the debugger, but as this is my first sparring with Deno, sadly to no great success.

Some pointers that may help, or could be red herrings: 

* The 204 requests used to fail with the error that is raised [here](https://github.com/denoland/deno/blob/6bea6b31d3b502165e7e890c889919d67df1a795/ext/fetch/23_response.js#L203)
* This error happens when `bodyWithType` is not null ([see here](https://github.com/denoland/deno/blob/6bea6b31d3b502165e7e890c889919d67df1a795/ext/fetch/23_response.js#L322))
* The `body` is created by hopping through a few layers of `webidl.converters`, which is where I lost track with my current debugger setup 😞 . From looking at the code in GitHub, there's a chance that `converters.DOMString` is called which could return `'null'` rather than `null`.
* Tried to reproduce with plain Deno (without supabase) but the function seems to have worked there. So I'm not sure whether Deno fixed it in their latest version already, or I chased down an incorrect assumption.

andreespirela:
Fixes #91


crates/base/tests/null_body_status_null_body_tests.rs
```diff
@@ -69,3 +71,21 @@ async fn test_null_body_with_204_status_post() {
 
     assert_eq!(body_bytes.len(), 0);
 }
+
+#[tokio::test]
+async fn test_custom_readable_stream_response() {
```
andreespirela, line 76: Chore: Integration test utilities to create these tests

andreespirela: For some other ticket

laktek (approved):

laktek:
Can you check why the tests are failing?

crates/sb_workers/user_workers.js
```diff
@@ -49,7 +50,9 @@ class UserWorker {
 
         const res = await core.opAsync("op_user_worker_fetch_send", this.key, requestRid);
         const bodyStream = readableStreamForRid(res.bodyRid);
-        return new Response(res.size ? bodyStream : null, {
+        const uint8ResBody = (await readableStreamCollectIntoUint8Array(bodyStream)).buffer;
```
TimBeyer, line 53: Wouldn't this now buffer any streamed responses? It seems like that would make it impossible to use this to develop anything where streaming of a response is required. 

Streaming worked fine before the 204 fix so maybe there's a way to preserve both things?

andreespirela: In the long run we want to be able to do this. I'd consider this a sort of quick fix while we figure some stuff out with Streamers, and the way the responses are shared through channels. We've discussed this, and even with Deno, there's cases where ReadableStream is also consumed internally when being returned. There's some challenges around this and it'll get to our list of priorities :)

PR #94, opened 2023-05-16
Remove worker after completion

laktek:
### correctly remove worker from pool 

Moved the logic to remove a worker from pool to worker_ctx, so it will
be done all instances. Previously, crashes would leave dead workers in
the pool.

### [reuse user workers by default](https://github.com/supabase/edge-runtime/commit/508ece44e3392688ced3678c7d0e3ced76da7863) 

This changes user workers to be reused by default, this can be overriden
by passing a `forceCreate` option when creating a worker.

Previously, main function had to maintain a cache of workers leading to
requests sent to already removed workers. This change would prevent
that.

### DIFF
--- a/crates/base/src/worker_ctx.rs
+++ b/crates/base/src/worker_ctx.rs
@@ -150,9 +150,9 @@ pub async fn create_user_worker_pool() -> Result<mpsc::UnboundedSender<UserWorke
                     let key = city_hash_64(key_input.as_bytes());
 
                     // do not recreate the worker if it already exists
-                    if let Some(_worker) = user_workers.get(&key) {
-                        // if the force create option is set skip the existing worker
-                        if !user_worker_rt_opts.force_create {
+                    // unless force_create option is set
+                    if !user_worker_rt_opts.force_create {
+                        if let Some(_worker) = user_workers.get(&key) {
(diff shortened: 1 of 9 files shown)


crates/base/src/worker_ctx.rs
```diff
@@ -112,18 +127,39 @@ pub async fn create_user_worker_pool() -> Result<mpsc::UnboundedSender<UserWorke
...
+                    let service_path = worker_options.service_path.to_str().unwrap_or("");
+                    let mut key_input = service_path.to_string();
+                    if user_worker_rt_opts.force_create {
+                        let cur_epoch_time = SystemTime::now().duration_since(UNIX_EPOCH)?;
+                        key_input = format!("{}-{}", key_input, cur_epoch_time.as_millis());
+                    }
+                    let key = city_hash_64(key_input.as_bytes());
+
+                    // do not recreate the worker if it already exists
+                    if let Some(_worker) = user_workers.get(&key) {
```
andreespirela, line 153: Why this check if `force_create` is enabled?

laktek: ah yeah, good point. Probably better to swap the two checks

andreespirela (approved):
lgtm

PR #105, opened 2023-05-26
feat: add support for providing import map as a data URI

laktek:

This PR adds the option to provide an import_map_path to a user worker as a data URI.


andreespirela (approved):
LGTM

PR #110, opened 2023-05-29
Function Events

andreespirela:
This PR introduces the creation and handling of events.

Summary: A new worker type now exists. Which is `EventWorker`. Think of this as an event manager in the background which is constantly receiving events without blocking the main service function **at all** or the user function. 

Permissions:
The Event Manager has access to all the same globals as the `Main` runtime with only one exception, **Event Manager cannot create User workers.**

TODO:

- [ ]  Implement all events
- [ ]  Discuss whether we want to dispatch an event (not necessary, but if we're sticking to certain kind of practices we might want to be opinionated about how the Event Manager is written as far as the typescript code)
- [ ] Tests
- [ ] Disable Event Ops and Code (such as `SupabaseEventListener`) from both `User` and `Main` runtime


crates/sb_workers/event_worker.js
```diff
@@ -0,0 +1,41 @@
...
+        } catch (e) {
+            console.log(e);
+        }
+    }
+
+    #buildEvent(event) {
+        const rawEvent = event['Event'];
+        const eventType = Object.keys(rawEvent)[0];
+        return {
+            deployment_id: "",
```
andreespirela, line 18: - [ ] Implement `DeploymentId`, `Region`, `ExecutionId`

examples/event-manager/index.ts
```diff
@@ -0,0 +1,7 @@
+const eventManager = new globalThis.EventManager();
+
+for await (const data of eventManager) {
```
andreespirela, line 3: Infinite loop

andreespirela: Well, until the receiver is dropped

crates/base/src/server.rs
```diff
@@ -87,24 +90,43 @@ impl Server {
...
     ) -> Result<Self, Error> {
+        let mut worker_event_sender: Option<mpsc::UnboundedSender<WorkerEvents>> = None;
+
+        // Creates Event Worker
+        if let Some(events_service_path) = maybe_events_service_path {
+            let events_path = Path::new(&events_service_path);
+            let events_path_buf = events_path.to_path_buf();
+
+            let event_worker =
+                create_event_worker(events_path_buf, import_map_path.clone(), no_module_cache)
```
andreespirela, line 106: It creates an event manager for all the the user function pool. So all functions will report to this instance

crates/base/src/server.rs
```diff
@@ -87,24 +90,43 @@ impl Server {
         ip: &str,
         port: u16,
         main_service_path: String,
+        maybe_events_service_path: Option<String>,
```
andreespirela, line 93: Maybe ?

crates/sb_workers/event_worker.js
laktek: Can we make these fields configurable via event worker?

crates/base/src/server.rs
laktek: since it's an option type maybe isn't really necessary.

laktek (approved):

PR #111, opened 2023-05-31
Uncaught exception, memory & time limit events

andreespirela:

Introduces events for:
- uncaughtException
- memoryLimit
- timeLimit

### DIFF
--- a/crates/base/src/edge_runtime.rs
+++ b/crates/base/src/edge_runtime.rs
@@ -20,6 +20,7 @@ use tokio::sync::mpsc;
 use tokio::sync::oneshot;
 use urlencoding::decode;
 
+use crate::utils::send_event_if_event_manager_available;
 use crate::{errors_rt, snapshot};
 use module_loader::DefaultModuleLoader;
 use sb_core::http_start::sb_core_http;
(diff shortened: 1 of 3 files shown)


crates/base/src/edge_runtime.rs
```diff
@@ -359,11 +373,13 @@ impl EdgeRuntime {
                 tokio::select! {
                     _ = tokio::time::sleep(Duration::from_millis(worker_timeout_ms)) => {
                         debug!("max duration reached for the worker. terminating the worker. (duration {})", human_elapsed(worker_timeout_ms));
+                        send_event_if_event_manager_available(event_sender, WorkerEvents::TimeLimit(PseudoEvent {}));
```
laktek, line 376: Should we move this into `worker_ctx`? So all events are sent from it?

andreespirela: The `tokio::select! {` ?

laktek: I meant the sending events. Since the result of the worker is sent back to the context, it's better to handle sending events from there.

Makes it easier to refactor in the future.

laktek: hmm..I thought we are returning the status in the `EdgeCallResult`? https://github.com/supabase/edge-runtime/blob/main/crates/base/src/edge_runtime.rs#L265

andreespirela: No yet. That’s one of the refactors I’ve been trying to make but it messes with the response.

laktek: alright, then lets go with the current approach and then refactor it later.

laktek (approved):

PR #117, opened 2023-06-16
chore: add names to worker threads

laktek:

Name the worker threads so it's easy to debug. Eg: `top -H -p pid` will show all threads run by edge runtime.
Screen Shot 2023-06-17 at 9 35 57 am


andreespirela (approved):
Lgtm

PR #123, opened 2023-06-23
Measure CPU time used by a worker

laktek:

This PR introduces way to measure CPU time used by user workers and enforce max CPU time use.

Also, includes some refactoring to move all enforcements into a supervisor thread.

### DIFF
--- a/crates/base/src/deno_runtime.rs
+++ b/crates/base/src/deno_runtime.rs
@@ -1,4 +1,4 @@
-use crate::utils::units::{bytes_to_display, human_elapsed, mib_to_bytes};
+use crate::utils::units::mib_to_bytes;
 
 use crate::js_worker::module_loader;
 use anyhow::{anyhow, Error};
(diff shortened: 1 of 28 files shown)


crates/base/src/deno_runtime.rs
```diff
@@ -207,8 +207,8 @@ impl EdgeRuntime {
                 }
             },
             get_error_class_fn: Some(&get_error_class_name),
-            shared_array_buffer_store: None,
-            compiled_wasm_module_store: None,
+            shared_array_buffer_store: Default::default(),
```
andreespirela, line 210: Do we need to set them?

andreespirela:
Great work

crates/base/src/deno_runtime.rs
laktek: Will enable only the `compiled_wasm_module_store`. It should speed up wasm usages across isolates.

andreespirela (approved):
LGTM

PR #127, opened 2023-06-27
fix: show more descriptive error when module not found

### LINKED ISSUE #126: Show exactly which import is failing
# Chore

Always great to see all the improvements on the edge runner! Hope we can do this DX improvement as well!

## Describe the chore

Make it more visible and descriptive when there is an import error.

## Additional context

I was importing another function from another file but forgot the `.ts` extension:

```typescript
import { parseAlert } from './parser' // insead of parser.ts
```

When I serve the function and call it I will get this error:

```
serving the request with /home/deno/functions/handle-resource
worker "/home/deno/functions/handle-resource" returned an error: No such file or directory (os error 2)
InvalidWorkerCreation: worker boot error
    at async Function.create (ext:sb_user_workers/user_workers.js:80:21)
    at async Server.<anonymous> (file:///home/deno/main/index.ts:95:24)
    at async Server.#respond (https://deno.land/std@0.182.0/http/server.ts:220:24)
```

I couldn't really tell what the error was. I misunderstood and thought the function folder was missing and wonder if some volume mounting would have caused the issue.

I think just displaying the line where the issue happened would be very helpful! Also seeing worker boot error feels very intimidating and distracted me from the fact that it was just a small typo

laktek:

Fixes #126 - Show a descriptive error message including the name of failed import.

### DIFF
--- a/crates/base/src/js_worker/module_loader.rs
+++ b/crates/base/src/js_worker/module_loader.rs
@@ -1,4 +1,4 @@
-use anyhow::{bail, Error};
+use anyhow::{anyhow, bail, Error};
 use deno_ast::EmitOptions;
 use deno_ast::MediaType;
 use deno_core::error::AnyError;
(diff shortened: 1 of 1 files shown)


andreespirela (approved):
LGTM

PR #130, opened 2023-07-03
fix: send console events from user workers to event worker

laktek:

Send console events from user worker to event worker for logging purposes.


crates/sb_core/js/bootstrap.js
```diff
@@ -244,18 +248,25 @@ globalThis.bootstrapSBEdge = (opts, isUserRuntime, isEventManager) => {
...
   setNumCpus(1); // explicitly setting no of CPUs to 1 (since we don't allow workers)
   setUserAgent("Supabase Edge Runtime");
   setLanguage("en");
 
-  ObjectDefineProperty(globalThis, "Deno", readOnly(denoOverrides));
 
-  if(isUserRuntime) {
-    loadUserRuntime();
+  if(isUserWorker) {
+    delete globalThis.EdgeRuntime;
```
laktek, line 259: @andreespirela I've inlined the `loadUserRuntime` here. It's handy to have it access to the scope here (so it can access `core` and `ops`.)

crates/sb_workers/event_worker.js
```diff
@@ -3,35 +3,33 @@ const { SymbolAsyncIterator } = primordials;
...
-            execution_id: "",
-            region: "",
-            context: undefined
-        }
-    }
-
     [SymbolAsyncIterator]() {
-        const scopedClass = this;
         return {
             async next() {
```
laktek, line 8: @andreespirela I simplified the symbol iterator a bit. Was there any specific reason we had the `nextEvent` function?

crates/sb_workers/events.rs
```diff
@@ -24,26 +22,25 @@ pub struct IncomingEvent {
...
         let mut op_state = state.borrow_mut();
-        op_state.take::<mpsc::UnboundedReceiver<WorkerEvents>>()
+        op_state.try_take::<mpsc::UnboundedReceiver<WorkerEvents>>()
     };
+    if rx.is_none() {
+        bail!("events worker receiver not available")
+    }
+    let mut rx = rx.unwrap();
 
-    let data = rx.try_recv();
```
laktek, line 33: @andreespirela Changed this to use async `recv` instead of `try_recv`. Previously, this was processing the empty values causing CPU spikes.

inian (approved):

PR #133, opened 2023-07-07
fix: set a module root for loading modules

laktek:

* Allow specifying a module root to load modules from
* Option to turn off loading remote modules


andreespirela (approved):
Lgtm!

PR #134, opened 2023-07-10
Add metadata to events

laktek:

This adds a `metadata` field to the event object. Currently it's populated with `service_path` and `execution_id`.


andreespirela (approved):
LGTM

PR #135, opened 2023-07-10
fix: add metadata to console events

laktek:

Console events logged from user functions will also include the metadata.


andreespirela (approved):
Lift off

PR #136, opened 2023-07-10
fix: print to stderr when logger is not available

laktek:

Print to stderr if no logger is defined.


andreespirela (approved):
LGTM

PR #141, opened 2023-07-23
Log source file

laktek:

If you run edge-runtime with `--log-source` option, it will emit log messages with the corresponding source file and line number.

eg:
```
crates/base/src/server.rs-164: client connection error (hyper::Error(IncompleteMessage))
```

This is quite useful for figuring out the source of some of the non-descriptive error messages.

### DIFF
--- a/crates/base/src/server.rs
+++ b/crates/base/src/server.rs
@@ -159,7 +159,9 @@ impl Server {
                                 .serve_connection(conn, service);
 
                              if let Err(e) = conn_fut.await {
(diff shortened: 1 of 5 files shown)


andreespirela:
@laktek should we close this one and have it in the other one instead?

laktek:
@andreespirela I just branched off from this PR. Good to merge this on its own. Can you approve?

laktek:
Merged as part of #142

PR #142, opened 2023-07-23
Option to disable net access in user workers

laktek:

Provides a new option to disable net access in the user worker.


andreespirela (approved):
LGTm

PR #143, opened 2023-07-25
fix: main function example

laktek:

Fix the main function example


andreespirela (approved):
LGTM

PR #145, opened 2023-07-26
fix: do not panic if unix rx is not available

laktek:

This should fix panics if unix channel receiver is not available.


andreespirela (approved):
nice, lgtm.

PR #151, opened 2023-07-31
fix: simplify Deno.osUptime implementation

laktek:

Simplifies `Deno.osUptime` to return the Edge Runtime uptime instead of underlying OS uptime.


andreespirela (approved):
I think it should be fine, go ahead. Lgtm

PR #152, opened 2023-07-31
fix: remove perl from runtime container

laktek:

There's a container vulnerability marked as high severity for the Perl module and edge-runtime doesn't rely on it.


andreespirela (approved):
Interesting! LGTM

PR #155, opened 2023-08-02
Refactor edge rt pt 2

andreespirela:
- Re-utilize code
- Change references to event_worker instead of event_manager


crates/cli/src/main.rs
```diff
@@ -85,7 +85,7 @@ fn main() -> Result<(), anyhow::Error> {
                     .cloned()
                     .unwrap();
                 let event_service_manager_path =
-                    sub_matches.get_one::<String>("event-manager").cloned();
+                    sub_matches.get_one::<String>("event-worker").cloned();
```
andreespirela, line 88: @laktek Note when updating **production**

laktek (approved):

PR #157, opened 2023-08-03
fix: parse unknown media type as JS

laktek:

Treat unknown media types as JS


andreespirela:
Lgtm

andreespirela (approved):

PR #158, opened 2023-08-10
feat: added sb_eszip extension

laktek:

This PR introduces a new extension to handle eszip extraction. It can be used from main function with the following API `EdgeRuntime.eszip.extract('/path/to/eszip', '/path/to/dest')`


andreespirela (approved):
Interesting. LGTM

PR #167, opened 2023-08-30
fix: globalThis.addEventListener

laktek:

Correctly sets the `Window.prototype` on globalThis.


andreespirela (approved):
lgtm

PR #169, opened 2023-09-03
feat: support loading inline module code

laktek:
Can provide module code as a string when creating a worker.

example,

```
const worker = createWorker({
....
maybeEntrypoint: "file:///src/index.ts",
maybeModuleCode: 'Deno.serve((req) => new Response("Hello from Module Code"));',
})
```

### DIFF
--- a/crates/base/src/deno_runtime.rs
+++ b/crates/base/src/deno_runtime.rs
@@ -119,10 +119,11 @@ impl DenoRuntime {
             conf,
             maybe_eszip,
             maybe_entrypoint,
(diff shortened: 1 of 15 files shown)


andreespirela (approved):
Lgtm!

adamcohenhillel:
This PR seems to be the one that broke my Deno Edge Function Environment - it stopped working on the 4th of September. Not sure what it's doing, as I'm not fully understand the codebase, but: this is the only PR and the release of the day it broke, and it relates to module code / loading

PR #170, opened 2023-09-05
feat: Implement Deno version according to latest upgrade

andreespirela:

This PR implements the latest `Deno.version` according to the latest upgrade we had on the crates which was Deno 1.36.0.

Note we append `supabase-edge-runtime` in order to differentiate.


laktek (approved):

laktek:
@andreespirela can you merge main to see if it will fix the fmt and clippy errors?

PR #173, opened 2023-09-12
feat: Option to abort user worker requests

laktek:

This PR adds the option to abort user worker requests.

Also, updated the implementation to correctly close the response bodies for null-body responses.


andreespirela (approved):
Good! Lgtm

PR #177, opened 2023-09-16
fix: option to allow remote modules in user workers

laktek:

Added the option to configure the option to allow remote modules in user workers.


andreespirela (approved):
Nice!

PR #179, opened 2023-09-18
fix: Made user worker create non-blocking

laktek:

Currently, the user worker creation blocks the worker pool events loops. This can cause deadlocks and make edge-runtime unresponsive. 

This PR modifies the user worker creation to be non-blocking.


andreespirela (approved):
Lgtm

PR #180, opened 2023-09-19
fix: exclude default v8 heap stats from user worker logs

laktek:

Excluding the v8 heap stats that were added by default to user worker logs. The op code is still there if we want to expose an explicity `console.memory` API.


andreespirela (approved):
Lgtm

PR #185, opened 2023-09-25
feat: Implement ESZIP factory for easier eszip building, and first st…

andreespirela:
Implement ESZIP factory for easier eszip building, and first steps towards NPM

### DIFF
--- a/crates/base/src/utils/graph_util.rs
+++ b/crates/base/src/utils/graph_util.rs
@@ -135,27 +135,23 @@ pub async fn build_graph_with_npm_resolution<'a>(
 
 pub async fn create_graph_and_maybe_check(
     roots: Vec<ModuleSpecifier>,
-) -> Result<Arc<deno_graph::ModuleGraph>, AnyError> {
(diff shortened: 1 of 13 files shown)


laktek:
I guess we can update the `bundle` command to take the input path as an arg and an optional `--output` path.

For example, it can be run as:

`./edge-runtime bundle --output dist/my_func.eszip src/my_func/index.ts`

crates/base/src/utils/graph_util.rs
```diff
@@ -0,0 +1,205 @@
...
+    // }
+
+    Ok(graph)
+}
+
+pub async fn create_module_graph_from_path(
+    main_service_path: &str,
+) -> Result<ModuleGraph, Box<dyn std::error::Error>> {
+    let main_service_directory = PathBuf::from(main_service_path);
+    let index = main_service_directory.clone().join("index.ts");
```
laktek, line 186: we can allow users to specify the full path to the entrypoint instead of the directory. That way we don't have to assume the entrypoint is always a `index.ts`

laktek (approved):

PR #186, opened 2023-09-27
fix: provide optional entrypoints for main/events eszips

laktek:

Added two CLI options `--main-entrypoint` and `--events-entrypoint` to provide entrypoints when eszips are used for main and events workers.

### DIFF
--- a/crates/base/src/macros/test_macros.rs
+++ b/crates/base/src/macros/test_macros.rs
@@ -26,7 +26,11 @@ macro_rules! integration_test {
                 None,
                 None,
                 false,
-                Some(tx.clone())
+                Some(tx.clone()),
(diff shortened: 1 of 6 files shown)


crates/base/src/macros/test_macros.rs
```diff
@@ -26,7 +26,11 @@ macro_rules! integration_test {
                 None,
                 None,
                 false,
-                Some(tx.clone())
+                Some(tx.clone()),
+                $crate::server::WorkerEntrypoints {
```
andreespirela, line 30: Is this right? `$crate`

laktek: yep, because it's in a macro

andreespirela (approved):
Lgtm

PR #187, opened 2023-10-03
fix: handle errors in Deno.serve

laktek:

Handle errors thrown inside handler functions in `Deno.serve`, so they don't break the event loop.


andreespirela (approved):
Lgtm, thanks!

PR #190, opened 2023-10-12
feat: include isolate memory & cpu usage in shutdwon event

laktek:

This PR introduces a new `WorkerEvents::Shutdown` which includes the reason, cpu time, and memory used by the isolate as metadata. The separate  `CpuTimeLimit`, `WallClockTimeLimit`, `MemoryLimit` worker events are removed in favor of this. Event worker implementations need to update to use the `Shutdown` event instead of previous events.

### DIFF
--- a/crates/base/src/rt_worker/worker.rs
+++ b/crates/base/src/rt_worker/worker.rs
@@ -5,7 +5,7 @@ use crate::utils::send_event_if_event_worker_available;
 use anyhow::{anyhow, Error};
 use cpu_timer::get_thread_time;
 use event_worker::events::{
-    EventMetadata, LogEvent, LogLevel, WorkerEventWithMetadata, WorkerEvents,
+    EventMetadata, ShutdownEvent, UncaughtExceptionEvent, WorkerEventWithMetadata, WorkerEvents,
 };
 use log::{debug, error};
 use sb_worker_context::essentials::{UserWorkerMsgs, WorkerContextInitOpts};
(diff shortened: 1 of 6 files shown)


crates/base/src/rt_worker/worker.rs
```diff
@@ -124,31 +121,36 @@ impl Worker {
                     }
                 });
 
+                let end_time = get_thread_time()?;
+                let cpu_time_used =
+                    usize::try_from((end_time - start_time) / 1_000_000).unwrap_or(0);
```
andreespirela, line 126: When might this unwrap to err so it equals 0 ? Wondering if this could be a vector attack somehow

crates/event_worker/events.rs
```diff
@@ -8,15 +8,35 @@ pub struct PseudoEvent {}
...
 
 #[derive(Serialize, Deserialize, Debug)]
-pub struct UncaughtException {
+pub struct IsolateMemoryUsed {
+    pub heap_used: usize,
+    pub external_memory: usize,
+}
+
+#[derive(Serialize, Deserialize, Debug)]
+pub enum ShutdownReason {
```
andreespirela, line 23: Amazing!

andreespirela (approved):
Lgtm!

crates/base/src/rt_worker/worker.rs
laktek: both start and times comes from syscall to get CPU times. It returns a int64, but we convert it to a usize as it cannot technically be less than a 0. Defaults to a 0 in case value cannot be converted to a valid uszie.

PR #191, opened 2023-10-12
fix: include total memory usage in shutdown event

laktek:

* removes unused memory usage op
* include total memory usage (heap + external) in shutdown event metadata


andreespirela (approved):
LGTM

PR #194, opened 2023-10-19
fix: release requires cmake

laktek:

The release fails since build require cmake. Install cmake and remove other unused files to free up space.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,6 +52,21 @@ jobs:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
 
+      - name: Prepare build environment
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          sudo docker image prune --all --force
+
+          sudo apt-get remove --purge -y man-db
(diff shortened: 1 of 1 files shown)


soedirgo:
Isn't cmake needed in the Dockerfile instead of the CI runner?

laktek:
> Isn't cmake needed in the Dockerfile instead of the CI runner?

@soedirgo ah yeah, good catch! fixed it in https://github.com/supabase/edge-runtime/pull/194/commits/3e4b28f17a8728809b5712cfe5645c8af53e66bb (need to remove other deps to reduce the disk space in the build container though)

.github/workflows/release.yml
```diff
@@ -52,6 +52,20 @@ jobs:
...
+      - name: Prepare build environment
+        run: |
+          sudo rm -rf /usr/share/dotnet
+          sudo rm -rf /usr/local/lib/android
+          sudo rm -rf /opt/ghc
+          sudo rm -rf /opt/hostedtoolcache/CodeQL
+          sudo docker image prune --all --force
+
+          sudo apt-get remove --purge -y man-db
+          sudo apt-get remove 'clang-13*' 'clang-14*' 'clang-15*' 'llvm-13*' 'llvm-14*' 'llvm-15*' 'lld-13*' 'lld-14*' 'lld-15*'
```
soedirgo, line 64: Are these to leave enough space for the Docker build to succeed? Otherwise since the runner runs these jobs in a container they should get cleaned up

soedirgo (approved):

PR #216, opened 2023-11-24
feat: set soft and hard limits for CPU time

laktek:
This PR modifies how the CPU time limits are enforced. Instead of accounting for bursts, we do set soft and a hard limit. When the soft limit is reached the worker is retired (ie. current request completes, but no new requests are accepted). And worker is terminated only when it hits the hard limit. This should reduce the chances of function failing mid-request due to CPU limits.


andreespirela (approved):
LGTM

stonediggity:
I've just worked through this video and the repo and am experiencing the same issue. I upgraded to the pro plan also as I was planning on using it for a production app but it's still shuts down as the CPU time limit is reached. I have also reduced the batch size to 1 and reduced the text chunk size to 500 in the process-markdown script.
Screenshot 2024-01-05 174646

PR #218, opened 2023-11-29
fix: load import map from eszip

laktek:

This PR contains a fix to correctly load import map from eszips.


laktek:
changes were incorporated into #217

PR #219, opened 2023-11-30
fix: expose execution_id as a environment variable

laktek:

It will be available to access within functions as `SB_EXECUTION_ID`.


andreespirela (approved):
Lgtm

PR #220, opened 2023-11-30
fix: Support module_code for eszipping code only

andreespirela:


crates/base/src/deno_runtime.rs
```diff
@@ -170,8 +173,16 @@ impl DenoRuntime {
             let arc_emitter_factory = Arc::new(emitter_factory);
 
             let main_module_url_file_path = main_module_url.clone().to_file_path().unwrap();
+
+            let maybe_code = if only_module_code {
```
laktek, line 177: can't we directly pass `maybe_module_code` as an option to `generate_binrary_eszip`

crates/sb_graph/graph_util.rs
```diff
@@ -303,11 +305,34 @@ pub async fn create_eszip_from_graph_raw(
...
-    let format_specifier = format!("file:///{}", specifier);
-    let module_specifier = ModuleSpecifier::parse(&format_specifier).unwrap();
+pub async fn create_graph(
+    file: PathBuf,
+    emitter_factory: Arc<EmitterFactory>,
+    maybe_code: &Option<FastString>,
+) -> ModuleGraph {
+    let module_specifier = if let Some(code) = maybe_code {
+        let specifier =
+            ModuleSpecifier::parse("http://internal.supabase.localhost/index.ts").unwrap();
```
laktek, line 315: can this be a file specifier like `file:///src/index.ts`?

laktek (approved):
2 minor comments. Otherwise LGTM.

crates/base/src/deno_runtime.rs
andreespirela: It's under `only_module_code` so that ```        let only_module_code =
            maybe_module_code.is_some() && maybe_eszip.is_none() && !is_some_entry_point;``` it only applies for this case.

PR #225, opened 2023-12-03
feat: enforce limits on array buffer allocations

laktek:

Switch to use a custom array buffer allocator, which can enforce limits on the max size of array buffer.


andreespirela (approved):
Lgtm

PR #228, opened 2023-12-06
fix: handle invalid paths to bundle command

laktek:

Show proper errors when invalid paths provided for the bundle command.


andreespirela (approved):
Thanks!

PR #227, opened 2023-12-06
fix: crash when V8 tries to JIT compilation in the PKU-supported environment

nyannyacha:
## What kind of change does this PR introduce?
Bug fix

## Description
This PR fixes the segfault when occurred the V8 tries to JIT compilation in running environments that supports PKU(Memory Protection Key; PKeys Userspace) feature.

It should only affect the environment that the PKU feature has enabled. You can check whether your environment supports the PKU feature using the below command:

```bash
lscpu | grep pku

# output
Flags:                              ... xsaveopt arat pku ospke avx512_vnni...
```

Fixes #199 and might be
Resolves #200, resolves #204, resolves #207, resolves #226

PR #234, opened 2023-12-14
refactor: set v8 flags before initialize v8 platform

nyannyacha:

Bug fix, Refactor

## Description

After #227 the platform is initialized explicitly at the entry point. However, after initializing the platform, it was no longer able to set the v8 flags [\[1\]][1] [\[2\]][2] [\[3\]][3], so it is necessary to adjust the invocation order of the set v8 flag function to avoid v8's static assertion.

[1]: https://source.chromium.org/chromium/chromium/src/+/main:v8/src/flags/flags.cc;l=1074;drc=8e78783dc1f7007bad46d657c9f332614e240fd8;bpv=1;bpt=1
[2]: https://source.chromium.org/chromium/chromium/src/+/main:v8/src/flags/flags.cc;drc=8e78783dc1f7007bad46d657c9f332614e240fd8;l=912
[3]: https://source.chromium.org/chromium/chromium/src/+/main:v8/src/init/v8.cc;l=261;drc=8e78783dc1f7007bad46d657c9f332614e240fd8;bpv=1;bpt=1

### DIFF
--- a/crates/base/src/commands.rs
+++ b/crates/base/src/commands.rs
@@ -1,6 +1,7 @@
 use crate::server::{Server, ServerCodes, WorkerEntrypoints};
 use anyhow::Error;
 use deno_core::JsRuntime;
+use log::error;
 use tokio::sync::mpsc::Sender;
 
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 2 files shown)


laktek (approved):

laktek:
Thanks for the contribution @nyannyacha (sorry for the delay in merging, I was on holidays)

PR #240, opened 2024-01-08
fix: Identified and fixed a memory leak from `JsRuntime`

### LINKED ISSUE #212: supabase functions serve runs out of memory and crashes with basic usage
**Describe the bug**
The edge-runtime does not terminate the worker for a few minutes after starting. This causes a pretty severe memory leak since a new worker is created for each request and never terminated: https://github.com/supabase/cli/blob/a0c8644deeef5a72f99687cc897eadc3dce256f1/internal/functions/serve/templates/main.ts#L135

This makes it effectively impossible to use `supabase functions serve` for local development.

**To Reproduce**
1. Start worker runtime
2. Send ~1000 requests within a few mins
3. Notice extremely high memory usage on the edge-runtime and crash.

**Expected behavior**
I expect the worker runtimes to be cleaned up after the request is complete if a new worker runtime is started for each request. Even nicer would be to reuse a single worker runtime and refresh it when files change, similar to `deno --watch`.

**Screenshots**
image
image

**Desktop (please complete the following information):**

- OS: macOS 14.0 (13-inch, M2, 2022)
- Version of supabase-js: beta (1.102.2)
- Version of Node.js: v20.5.0

**Additional context**
Disabling `forceCreate` solves the crashing issue, but breaks auto-reload.

I filed an issue on the edge-runtime here since I'm not sure which should be fixed to resolve the crashing problem. https://github.com/supabase/edge-runtime/issues/192

nyannyacha:

Bug fix

## Description

This PR fixes the memory leakage from `JsRuntime`.

I've already written some commits to the upstream fork to fix the memory leak problem.

The PR brings in that fork using the cargo patch[^1][^2].

CAVEATS: PR makes `edge-runtime` depend on the my upstream fork. If you have a plan that will manage upstream changes manually, it would make sense for the Supabase team to take my fork to your org repo first.

Resolves #212 

[^1]: https://github.com/nyannyacha/rusty_v8/tree/rokat
[^2]: https://github.com/nyannyacha/deno_core/tree/222

cc @laktek 
😋

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[env]
+RUSTY_V8_MIRROR = "https://github.com/nyannyacha/rusty_v8/releases/download"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -984,8 +984,7 @@ dependencies = [
 [[package]]
 name = "deno_core"
 version = "0.222.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b13c81b9ea8462680e7b77088a44fc36390bab3dbfa5a205a285e11b64e0919c"
+source = "git+https://github.com/nyannyacha/deno_core?branch=222#fd51ad55de7f24ac450af31a429a9902e40ad4b1"
 dependencies = [
  "anyhow",
  "bytes",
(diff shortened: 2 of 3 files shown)


laktek:
Have you opened PRs for `deno_core` and `rusty_v8`? We would have to stick to official releases due to supply chain security. We can merge this once those changes are pushed to upstream.

PR #241, opened 2024-01-08
feat: introduce various supervisor policy

### LINKED ISSUE #103: Error: channel closed
# Bug report

- [ ] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

image

```
ab -n 5000 -c 100 http://localhost:9000/empty-response
```

```
serving the request with ./examples/empty-response
Error: channel closed
    at async Function.create (ext:sb_user_workers/user_workers.js:80:21)
    at async createWorker (file:///Users/shijinhua/Documents/privatespace/Bun/github.com/mivim-runtime/runtime/examples/main/index.ts:33:16)
    at async callWorker (file:///Users/shijinhua/Documents/privatespace/Bun/github.com/mivim-runtime/runtime/examples/main/index.ts:48:28)
serving the request with ./examples/empty-response

```

Any other request will report an error after reporting an error

## To Reproduce

1.  `const workerTimeoutMs = 1 * 1000`
2.  cargo run -- start
3.  ab -n 5000 -c 100 http://localhost:9000/empty-response

## Screenshots

image
image

## System information

- OS: macOS
- Browser (if applies)  command line
- Version of none
- Version of Node.js: none

### LINKED ISSUE #192: forceCreate does not clean up isolates after they are created
# Bug report

### Describe the bug

Edge runtime always crashes after a few minutes of development. This is a huge blocker to using Supabase edge functions locally.

I've tried to gather all of the information I can, but I'm not super familiar with Rust async code.

I'm on an M2 MacBook Pro running edge-runtime through Docker.

## To Reproduce

1. Run this basic Oak server code with `supabase functions server --debug`:

```typescript
import { Application } from 'oak';
const app = new Application();
app.use((ctx) => {
  ctx.response.body = 'Hello world!';
});
await app.listen({ port: 8000 });
```

2. Send a request
3. Notice how edge-runtime times out and prints error logs instead of exiting gracefully
4. Using `autocannon` load testing tool: `autocannon -c 20 -d 60 localhost:54321/functions/v1/api/lol`
5. Notice how edge-runtime runs out of memory and crashes

## Expected behavior

Using Oak should exit gracefully after serving a request instead of timing out to free up resources. Or maybe reuse the isolate for other requests?

## System information

- OS: macOS 14.0 (13-inch, M2, 2022)
- Version of supabase-js: beta (1.102.2)
- Version of Node.js: v20.5.0

### Isolates never seem to be terminated

For every request, I see the following 5 log lines. From what I can tell, these mean that the isolate is timing out after a few seconds of CPU time/wall clock time. Shouldn't these only handle one request and then exit?

```
# oak server is timing out
# this happens for every request before eventually crashing
wall clock duration reached. isolate: 6e4779a6-4a69-499d-81d3-8af6e727464d
event loop error: Uncaught Error: execution terminated
DEBUG failed to send cpu alarm to the provided channel
DEBUG CPU time used: 160ms
DEBUG Sending warning alert CloseNotify
```

Also, I saw this strange case where the CPU time was negative. Maybe this is a hint at the underlying problem:
```
wall clock duration reached. isolate: 2dd53cc0-b8cd-4437-974c-c09313dec8ad
DEBUG failed to send cpu alarm to the provided channel
DEBUG CPU time used: -662ms
```

### Memory leak

I'm fairly sure this would cause a memory leak with many requests at the same time. I tried this command with the above code and saw a huge spike in memory and eventually a crash due to Docker killing the container:

```
autocannon -c 20 -d 60 localhost:54321/functions/v1/api/lol
```

Result:
```
Req/Bytes counts sampled once per second.
# of samples: 60

339 2xx responses, 4915 non 2xx responses
5k requests in 60.08s, 1.53 MB read
89 errors (74 timeouts)
```

CPU/Memory graph (only two samples as the test was starting since it crashed soon after):

SCR-20231014-qdhm

### LINKED ISSUE #197: Workers are reused to serve multiple requests but terminated on timeout
# Bug report
- [X] I confirm this is a bug with Supabase, not with my own application. (the issue even happens with the example "hello" function provided by supabase)
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

From my understanding, when a worker is spawned it always lives for `workerTimeoutMs` to serve requests and will then be terminated.
This causes functions executing on a previously started worker to randomly fail when the timout happens while they are executed.

## To Reproduce

- Set `workerTimeoutMs` in `index.ts` to some value like 10 seconds
- Set up a simple longer running function like this:
  ```ts
  serve(async (req) => {
    await new Promise(r => setTimeout(r, 5_000));
    return new Response(`"Hello from Edge Functions!"`, { headers: { "Content-Type": "application/json" } });
  });
  ```
- Repeatedly call this function
- When a request is started eg. 8 seconds after the first one, it will crash with the following log

```
2023-10-28 01:20:45 main function started
2023-10-28 01:21:24 [Info] Listening on http://localhost:9999/
2023-10-28 01:21:24 [Info] start
2023-10-28 01:21:32 [Info] end
2023-10-28 01:21:33 [Info] start
2023-10-28 01:21:34 wall clock duration reached. isolate: 9517434353123966400
2023-10-28 01:21:34 failed to send request to user worker: connection closed before message completed
2023-10-28 01:21:34 Error: InvalidWorkerResponse: user worker failed to respond
2023-10-28 01:21:34     at async UserWorker.fetch (ext:sb_user_workers/user_workers.js:52:15)
2023-10-28 01:21:34     at async Server.<anonymous> (file:///home/deno/functions/main/index.ts:79:16)
2023-10-28 01:21:34     at async Server.#respond (https://deno.land/std@0.131.0/http/server.ts:219:24)
```

## Expected behavior

Either workers should not be reused to serve multiple requests or their timeout should reset if they receive a new request.

PS: Also, there does not seem to be a way to adjust allowed CPU-time which is also a dealbreaker for long running workers like generating AI embeddings.

### LINKED ISSUE #198: unexpected worker error Invalid argument (os error 22)
# Bug report

- [ x] I confirm this is a bug with Supabase, not with my own application. 
- [ x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

After building edge-runtime with cargo I try to run an example and I get this error:

`RUST_BACKTRACE=full ~/edge-runtime-target/debug/edge-runtime start --main-service ./examples/main`

```

main function started
serving the request with ./examples/serve
unexpected worker error Invalid argument (os error 22)
TypeError: request body receiver not connected (request closed)
    at Object.write (ext:deno_web/06_streams.js:1113:20)
    at Module.invokeCallbackFunction (ext:deno_webidl/00_webidl.js:974:16)
    at WritableStreamDefaultController.writeAlgorithm (ext:deno_web/06_streams.js:3849:14)
    at writableStreamDefaultControllerProcessWrite (ext:deno_web/06_streams.js:4351:55)
    at writableStreamDefaultControllerAdvanceQueueIfNeeded (ext:deno_web/06_streams.js:4254:5)
    at writableStreamDefaultControllerWrite (ext:deno_web/06_streams.js:4398:3)
    at writableStreamDefaultWriterWrite (ext:deno_web/06_streams.js:4545:3)
    at Object.chunkSteps (ext:deno_web/06_streams.js:2693:17)
    at readableStreamFulfillReadRequest (ext:deno_web/06_streams.js:2535:17)
    at readableByteStreamControllerCommitPullIntoDescriptor (ext:deno_web/06_streams.js:2226:5)
```

## To Reproduce

1. Clone projet
2. Run cargo build in WSL 1
3. Get error message

## System information

- OS: Windows
- Version of supabase-js: github.com/supabase/edge-runtime tag v1.22.3
- Version of Node.js: n/a

## Additional context

I'm building edge runtime inside WSL 1, Ubuntu 22

nyannyacha:
Feature

## Description

This PR introduces various supervisor policies.

1. `per_worker`
The behavior of this policy same as the behavior before introducing this PR. It accumulates CPU time and Wall-clock time for each incoming request until hitting the specified cap.

2. `per_request`
It calculates CPU time per request and the Wall-clock time limit is calculated based on the last request. However, it terminates the isolate if either incoming request reaches a hard CPU time limit. This policy allows the isolation can exist for longer periods so it can maximize the benefits of JIT compilation.

3. `oneshot`
Development purpose only. This policy forces the supervisor to terminate the isolation immediately if the request is complete. It is most useful in the development environment. Because the isolate lives only until a request is done, developers can receive feedback easily about per source code change.

These policies can be specified through the CLI argument and in some cases, you can specify additional arguments that specified policy supports.

### Additional policy arguments

1. `max-parallelism` is the maximum count of workers that can exist in the worker pool simultaneously. It is calculated _per service path_. **(It applied to all policies)**

2. `request-wait-timeout` is only affected if using `per_request` policy. It represents a maximum time in milliseconds that can wait to establish a connection with a worker. It is necessary because the `per_request` supervisor will not allow requests if the worker is still working on the previous request. **(It only applied to `per_request`, `oneshot` policy)**

## Considerations of the `per_request` policy

As described earlier, In the `per_request` policy, one or more requests can't coexist in the isolate during unit time.

Thus, depending on the case, if the handler does not handle the requests fast enough, It might lead to consequences of increased latency or creating massive extra isolates.

So It's necessary to limit the active user workers count by specifying an appropriate `max-parallelism` on the CLI to prevent excessive memory consumption.

These characteristics show us that it's not a complete alternative to the `per_worker` policy, and there is a trade-off clearly between the two policies.

per_request drawio (2)[^1]

### Pooler will strictly limit the number of Active Isolates

This PR changes the behavior of the pooler to minimize memory resource consumption due to creating an isolate.

Previously, The pooler there was no limit to the active isolation count explicitly. These results appeared to unbalance memory usage per service path. Furthermore, allowing to create the unlimited isolation could appear as OOM, depending on the cases.

PR makes the pooler limit the active isolation count per service path by using the semaphore.

Active isolation count can be adjusted by specifying `--max-parallelism` in the CLI argument. It follows the host CPU(or core) count by default if the user has not specified it.

Additionally, on the `per_worker` policy, active workers going to a retired state shouldn't affect the active isolation count. That means even if all active workers go to a retired state at once, it MUST not affect the request throughput or make the pending requests as the orphan state for a long time.

### Each supervisor task is no longer assigned on its native thread

Creating the native threads is expensive, and considering the isolation could be created or removed massively, it is better to spawn the supervisor tasks as a green thread managed by the tokio runtime to minimize resource consumption.

### `per_worker` behavior change

If workers reach the retired state, they can't accept subsequent requests, but they exist until the wall clock timer has timed out, so it will cause the consumption of massive resources in some cases.

This PR mitigates this problem by implementing the early termination of retired workers by adding a routine tracking the lifetime of each request to the supervisor.

### The synchronized connection lifetime of each unix stream pair

Before this PR, each unix stream pair used to process requests had a different connection lifetime. These connection lifetime mismatch issue led to an unexpected EOFs, and sometimes to internal server errors or timeouts.

This PR fixes the problem by introducing a routine that preserves the lifetime of all unix stream pairs used in a request until the main thread flushes the response to the requester completely. This resulted in some performance regression in request throughput for the `per_worker` policy.

### CPU timer can be turned off optionally

This PR makes the user turn off the CPU timer limit optionally by specifying a zero for each CPU timer parameter.
```typescript
...
const cpuTimeSoftLimitMs = 0;
const cpuTimeHardLimitMs = 0;

return await EdgeRuntime.userWorkers.create({
    ...
    cpuTimeSoftLimitMs,
    cpuTimeHardLimitMs,
...
```

### Added `applyConnectionWatcher` function into `EdgeRuntime` namespace

It was introduced to address synchronization issues with Unix stream pair connection lifetimes described above.

This is to keep track of the connection lifetime to exactly which response object is associated with the cloned request object, and it should be called after when the `Request.clone()` method is used.

If you do not pass a cloned request object to this function, connection lifetime is not guaranteed. So this can lead to previous behavior like intermittent EOFs, internal server errors, or timeouts. (for backward compatibility, `Worker.fetch` will emit console warnings instead of hard errors to the cloned request object that was not passed to that function)

> Note: It's only used when the _user needs to clone the request_. If the request does not need to be cloned for any reason, invoking this function is unnecessary. Because the original request passed by an argument of the `serve` function is already assigned the reference for the connection watcher.

> Note: This is only relevant in the main worker context, not the user worker context.

```typescript

Object.defineProperty(globalThis, 'EdgeRuntime', {
	get() {
		return {
			userWorkers: SUPABASE_USER_WORKERS,
                         // 👇
			applyConnectionWatcher: (src, dest) => {
				applyWatcherRid(src, dest); 
			}
		};
	},
	configurable: true,
});

// Copies the connection watcher reference of `srcRequest` to `destRequest`.
// Throws `TypeError` if both arguments are not `Request`.
EdgeRuntime.applyConnectionWatcher(srcRequest: Request, destRequest: Request): void;
```

Usage
```typescript
// NOTE: It covers the case if the user NEEDS to clone the request.
// If you don't need to clone the request, then you don't have to call.
Deno.serve(async (req: Request) => {
	// clone the request
	const req2 = new Request(req); // 👈 or `req.clone()`

	// do something...

	const worker = await createWorker();
	const controller = new AbortController();

	// ...

	EdgeRuntime.applyConnectionWatcher(req, req2); // 👈
	return await worker.fetch(req2, { signal });

	// ...
});
```

## Benchmark[^2]

Running environment
```
Hardware: Macbook Air M1 2020, 16GB
Container: Docker/Colima(0.6.7, vz, no rosetta 2)
Guest OS: Debian/Bookworm
ARCH: aarch64/arm64
VCPU: 8
RAM: 8GiB (16GiB Swap)
```

example.json
```json
{
    "name": "meow"
}
```

examples/hello-world/index.ts
```typescript
Deno.serve((req: Request) => {
	const result = mySlowFunction(2);
	const data = {
		message: `Hello33 ${result}`,
		test: 'foo',
	};

	return Response.json(data);
});

function mySlowFunction(baseNumber) {
	// console.time('mySlowFunction');
	let result = 0;	
	for (var i = Math.pow(baseNumber, 7); i >= 0; i--) {		
		result += Math.atan(i) * Math.tan(i);
	};
	// console.timeEnd('mySlowFunction');
	return result;
}
```

### v1.30.0 vs PR-241 / per_worker / Normal
Limits
```typescript
const workerTimeoutMs = 20 * 1000;

// XXX(Nyannyacha): What CPU time limit is used in production on Supabase?
const cpuTimeSoftLimitMs = 1000 * 100;
const cpuTimeHardLimitMs = 1000 * 200;
```

command
```
hey -c 20 -z 10s -m POST -D ./example.json http://127.0.0.1:8888/hello-world
```

v1.30.0 (VCPU: 8, Max parallelism: _unspecified_)
```
Summary:
  Total:	10.0026 secs
  Slowest:	0.2269 secs
  Fastest:	0.0011 secs
  Average:	0.0044 secs
  Requests/sec:	4579.3274


Response time histogram:
  0.001 [1]	|
  0.024 [45723]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.046 [41]	|
  0.069 [12]	|
  0.091 [8]	|
  0.114 [0]	|
  0.137 [0]	|
  0.159 [0]	|
  0.182 [9]	|
  0.204 [1]	|
  0.227 [10]	|


Latency distribution:
  10% in 0.0026 secs
  25% in 0.0031 secs
  50% in 0.0038 secs
  75% in 0.0052 secs
  90% in 0.0064 secs
  95% in 0.0071 secs
  99% in 0.0096 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0000 secs, 0.0011 secs, 0.2269 secs
  DNS-lookup:	0.0000 secs, 0.0000 secs, 0.0000 secs
  req write:	0.0000 secs, 0.0000 secs, 0.0052 secs
  resp wait:	0.0043 secs, 0.0011 secs, 0.2236 secs
  resp read:	0.0001 secs, 0.0000 secs, 0.0135 secs

Status code distribution:
  [200]	45777 responses
  [500]	28 responses
```

PR-241 (VCPU: 8, Max parallelism: 8 by default)
```
Summary:
  Total:	10.0047 secs
  Slowest:	0.1230 secs
  Fastest:	0.0014 secs
  Average:	0.0046 secs
  Requests/sec:	4361.3648


Response time histogram:
  0.001 [1]	|
  0.014 [43483]	|■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■
  0.026 [102]	|
  0.038 [8]	|
  0.050 [0]	|
  0.062 [1]	|
  0.074 [2]	|
  0.087 [17]	|
  0.099 [0]	|
  0.111 [2]	|
  0.123 [18]	|


Latency distribution:
  10% in 0.0029 secs
  25% in 0.0033 secs
  50% in 0.0040 secs
  75% in 0.0055 secs
  90% in 0.0068 secs
  95% in 0.0075 secs
  99% in 0.0097 secs

Details (average, fastest, slowest):
  DNS+dialup:	0.0000 secs, 0.0014 secs, 0.1230 secs
  DNS-lookup:	0.0000 secs, 0.0000 secs, 0.0000 secs
  req write:	0.0000 secs, 0.0000 secs, 0.0088 secs
  resp wait:	0.0045 secs, 0.0012 secs, 0.1201 secs
  resp read:	0.0001 secs, 0.0000 secs, 0.0134 secs

Status code distribution:
  [200]	43634 responses
```

### v1.30.0 vs PR-241 / per_worker / Chunked Request (Transfer-Encoding: chunked)
Limits
```typescript
const workerTimeoutMs = 20 * 1000;
const cpuTimeSoftLimitMs = 2000;
const cpuTimeHardLimitMs = 3000;
```

payload.bin
```
POST http://127.0.0.1:8888/hello-world
Content-Type: application/json
@./example.json
```

command
```
 cat payload.bin | vegeta attack -rate=1000 -duration=13s -chunked
```

v1.30.0
```
Requests      [total, rate, throughput]         13000, 1000.08, 991.46
Duration      [total, attack, wait]             13s, 12.999s, 1.136ms
Latencies     [min, mean, 50, 90, 95, 99, max]  911.917µs, 3.521ms, 1.182ms, 1.733ms, 4.452ms, 80.947ms, 253.677ms
Bytes In      [total, mean]                     690527, 53.12
Bytes Out     [total, mean]                     299000, 23.00
Success       [ratio]                           99.15%
Status Codes  [code:count]                      200:12889  500:111
Error Set:
500 Internal Server Error
```

PR-241
```
Requests      [total, rate, throughput]         13000, 1000.08, 999.99
Duration      [total, attack, wait]             13s, 12.999s, 1.2ms
Latencies     [min, mean, 50, 90, 95, 99, max]  949.625µs, 2.6ms, 1.258ms, 1.663ms, 3.17ms, 43.234ms, 158.649ms
Bytes In      [total, mean]                     689000, 53.00
Bytes Out     [total, mean]                     299000, 23.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:13000
Error Set:
```

----

Resolves supabase/supabase#19815
Resolves supabase/cli#247
Resolves #197
Resolves #198
Resolves #192 
Resolves #103 

[^1]: Figure is a briefly expressed flow chart of how the `per_request` policy works.
[^2]: `hey` was run in the same host, so the request per sec may not be accurate


nyannyacha:
```
error: use of a fallible conversion when an infallible one could be used
   --> crates/node/ops/crypto/mod.rs:413:65
    |
413 |                     .verify_prehash(digest, &signature.to_vec().try_into()?)
    |                                                                 ^^^^^^^^ help: use: `into`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_fallible_conversions
    = note: `-D clippy::unnecessary-fallible-conversions` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::unnecessary_fallible_conversions)]`
```

🧐

nyannyacha:
```
running 2 tests
test test_file_upload ... ok
error: test failed, to rerun pass `-p base --test oak_user_worker_tests`

Caused by:
  process didn't exit successfully: `/home/runner/work/edge-runtime/edge-runtime/target/debug/deps/oak_user_worker_tests-f0e7ee40c14d05fb` (signal: 14, SIGALRM: alarm clock)
Error: Process completed with exit code 101.
```

🙄

nyannyacha:
Is cargo clippy check running above 1.75.0?
Looking at that lint description, it introduced from 1.75.0
🤔

nyannyacha:
```
...
test deno_runtime::test::test_read_file_user_rt ... ok

failures:

error: test failed, to rerun pass `-p base --lib`
---- deno_runtime::test::test_create_eszip_from_graph stdout ----
thread 'deno_runtime::test::test_create_eszip_from_graph' panicked at crates/base/src/deno_runtime.rs:497:14:
called `Result::unwrap()` on an `Err` value: Reading /home/runner/.cache/deno/npm/registry.npmjs.org/is-even/1.0.0/.deno_sync_lock

Caused by:
    No such file or directory (os error 2)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    deno_runtime::test::test_create_eszip_from_graph

test result: FAILED. 10 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.53s
```

🧐

nyannyacha:
🫠

nyannyacha:
Note: This PR does not include the memory leak fixes.

nyannyacha:
rebased

crates/base/src/rt_worker/supervisor/strategy_per_request.rs
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            Some(notify) = req_start_rx.recv() => {
+                // INVARIANT: This branch MUST not be satisfied more than once
+                // during the same request cycle.
+                assert!(!req_start_ack, "supervisor has seen request start signal twice");
+
+                notify.notify_one();
+
+                if let Some(x) = cpu_timer.as_ref() {
```
andreespirela, line 58: Awesome PR! Could we have a more descriptive name for `x` ?

crates/base/src/rt_worker/supervisor/strategy_per_request.rs
```diff
@@ -0,0 +1,118 @@
...
+                if let Some(tx) = pool_msg_tx.clone() {
+                    if tx.send(UserWorkerMsgs::Idle(key)).is_err() {
+                        error!("failed to send idle msg to pool: {:?}", key);
+                    }
+                }
+
+                continue;
+            }
+
+            Some(x) => {
```
andreespirela, line 103: Same here

crates/base/src/rt_worker/supervisor/strategy_per_worker.rs
```diff
@@ -0,0 +1,144 @@
...
+        tokio::select! {
+            Some(_) = cpu_alarms_rx.recv() => {
+                if !cpu_time_soft_limit_reached {
+                    // retire worker
+                    is_retired.raise();
+                    error!("CPU time soft limit reached. isolate: {:?}", key);
+                    cpu_time_soft_limit_reached = true;
+
+                    if req_ack_count == demand.load(Ordering::Acquire) {
+                        let interrupt_data = IsolateInterruptData {
```
andreespirela, line 58: This struct seems like it can go outside, it's also re-used during the `else`

crates/base/src/rt_worker/supervisor/strategy_per_worker.rs
```diff
@@ -0,0 +1,144 @@
...
+                            error!("failed to send idle msg to pool: {:?}", key);
+                        }
+                    }
+                }
+
+                if !cpu_time_soft_limit_reached || req_ack_count != demand.load(Ordering::Acquire) {
+                    continue;
+                }
+
+                let interrupt_data = IsolateInterruptData {
```
andreespirela, line 94: Wondering if there's a way to move this out as it seems repeated, and maybe clone it as needed

crates/base/src/rt_worker/supervisor/strategy_per_worker.rs
nyannyacha: Yes, it's probably reusable.
In the case of `per_worker`, It just nothing but moved into the separated module, so I'm not touched aggressively before.

I'll fix it

crates/base/src/rt_worker/supervisor/strategy_per_request.rs
nyannyacha: >

> How's this line working also? I was getting an error, because `reset` is not part of CPUTimer, because it does not contain `&self`.
> 
> had to add
> 
> ```rust
> #[cfg(not(target_os = "linux"))]
>     pub fn reset(&self) -> Result<(), Error> {
>         Ok(())
>     }
> ```
> 
> (&self) for it to compile

Ah, that was probably my mistake.
As for the CPU timer, it's only available on Linux, so I hadn't actively looked at other platforms behind the feature flag.

I'll fix it

nyannyacha: solved

crates/base/src/rt_worker/supervisor/strategy_per_worker.rs
nyannyacha: I think it could be solved by using the `FnOnce` closure 😋

nyannyacha: solved

nyannyacha:
```
test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.28s

     Running tests/oak_user_worker_tests.rs (target/debug/deps/oak_user_worker_tests-f0e7ee40c14d05fb)

running 2 tests
test test_oak_server ... ok
error: test failed, to rerun pass `-p base --test oak_user_worker_tests`

Caused by:
  process didn't exit successfully: `/home/runner/work/edge-runtime/edge-runtime/target/debug/deps/oak_user_worker_tests-f0e7ee40c14d05fb` (signal: 14, SIGALRM: alarm clock)
Error: Process completed with exit code 101.
```

🙄

nyannyacha:
🤔
I have changed to turn off the CPUTimer-related bits on unit testing, I have no idea why `SIGALRM` triggered

nyannyacha:
Characteristics (`per_worker` vs `per_request`) / 3 min
----
Limits
```
const workerTimeoutMs = 20 * 1000;
const cpuTimeSoftLimitMs = 500;
const cpuTimeHardLimitMs = 600;
```
Command
```
vegeta attack -rate=0 -duration=3m -max-workers=12 [-chunked]
```

graph


### per_worker
```
Requests      [total, rate, throughput]         708556, 3936.39, 3936.36
Duration      [total, attack, wait]             3m0s, 3m0s, 1.341ms
Latencies     [min, mean, 50, 90, 95, 99, max]  911.042µs, 2.956ms, 2.566ms, 4.816ms, 5.462ms, 7.062ms, 144.479ms
Bytes In      [total, mean]                     37553468, 53.00
Bytes Out     [total, mean]                     16296788, 23.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:708556
Error Set:
```

### per_request
```
Requests      [total, rate, throughput]         715516, 3975.08, 3975.04
Duration      [total, attack, wait]             3m0s, 3m0s, 1.792ms
Latencies     [min, mean, 50, 90, 95, 99, max]  1.003ms, 2.958ms, 2.498ms, 5.002ms, 5.627ms, 7.152ms, 131.394ms
Bytes In      [total, mean]                     37922348, 53.00
Bytes Out     [total, mean]                     16456868, 23.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:715516
```

### per_worker / chunked
```
Requests      [total, rate, throughput]         661733, 3676.28, 3676.24
Duration      [total, attack, wait]             3m0s, 3m0s, 1.953ms
Latencies     [min, mean, 50, 90, 95, 99, max]  1.088ms, 3.22ms, 2.8ms, 5.071ms, 5.7ms, 7.533ms, 120.342ms
Bytes In      [total, mean]                     35071849, 53.00
Bytes Out     [total, mean]                     15219859, 23.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:661733
Error Set:
```

### per_request / chunked
```
Requests      [total, rate, throughput]         688821, 3826.77, 3826.74
Duration      [total, attack, wait]             3m0s, 3m0s, 1.694ms
Latencies     [min, mean, 50, 90, 95, 99, max]  999.458µs, 3.109ms, 2.65ms, 4.974ms, 5.672ms, 7.618ms, 110.429ms
Bytes In      [total, mean]                     36507513, 53.00
Bytes Out     [total, mean]                     15842883, 23.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:688821
Error Set:
```

nyannyacha:
Considerations of the `per_request` policy
---

As described earlier, In the `per_request` policy, one or more requests can't coexist in the isolate during unit time.

Thus, depending on the case, if the handler does not handle the requests fast enough, It might lead to consequences of increased latency or creating massive extra isolates.

So It's necessary to limit the active user workers count by specifying an appropriate `max-parallelism` on the CLI to prevent excessive memory consumption.

These characteristics show us that it's not a complete alternative to the `per_worker` policy, and there is a handoff clearly between the two policies.

per_request drawio[^1]

[^1]: Figure is a briefly expressed flow chart of how the `per_request` policy works.

andreespirela:
So I tested everything, it seems like all it's good to go. But before, I approve, it would be good if we can get some tests that touches on the different supervisors, and the limits and all. Like an integration test.

cc @nyannyacha 

Thanks for the work!

nyannyacha:
@andreespirela 
Ah, that's funny 🤣  I was just exploring that topic at the right time.

As a continuation of that exploration, I'm refactoring in the [other branch](https://github.com/nyannyacha/edge-runtime/tree/perf-use-green-thread) the native thread spawning parts that are used to run the runtime event loop as that can be used in the green thread pool. 😁

nyannyacha:
When you say integration testing, are you referring to integration testing in the context of rust, and if so, can you be more specific about what you need?

I can't think of anything right now 😅

andreespirela:
@nyannyacha Yes so we have a couple of integration tests you can take a look at for example, `oak_user_worker_tests.rs`, they are testing some of the behaviors of the calls for example. Here we want to test the behavior of the supervisors, just making sure the isolates get killed and everything that goes on when the supervisor signals take place.

nyannyacha:
@andreespirela 
Hi! 😋

Currently, I'm just refactoring the integration tests to be compatible with the new supervisor policies in the green thread PR.

I've added some flow control routines to the integration tests to reproduce the exact behaviors of these policies.

A little later, I'll submit a few commits for the integration test changes in the green thread PR.
Can you review and leave some comments for integration test changes?

andreespirela:
Yep will review ! @nyannyacha Thanks

nyannyacha:
Good! I'll submit changes after testing on the local machine!

I'll be happy if you would leave the comments for the integration test changes to better determine the direction of any integration tests that may written in the future.

laktek (approved):
Thanks for this excellent contribution.

spence:
awesome @nyannyacha! can you help migrate to the latest `deno_core` 😅

nyannyacha:
Hi! @spence 😋

Well... I'm not a team member of Supabase, so I think any issues with core dependencies are probably out of my hands.

nyannyacha:
I guess we could create a PR if they decide to, I don't know right now.

andreespirela:
@nyannyacha it’s okay! I’m working on migrating to newest Deno. Will keep you posted as far as this PR, there shouldn’t be any big issue

PR #246, opened 2024-01-22
fix: make major worker tasks managed separately from the user worker

nyannyacha:
Fix

## Description

This change makes major(main, event) workers spawn in a separate pool, not the user worker pool.

### DIFF
--- a/crates/base/src/rt_worker/rt.rs
+++ b/crates/base/src/rt_worker/rt.rs
@@ -8,7 +8,13 @@ pub static SUPERVISOR_RT: Lazy<tokio::runtime::Runtime> = Lazy::new(|| {
         .unwrap()
 });
 
-pub static WORKER_RT: Lazy<tokio_util::task::LocalPoolHandle> = Lazy::new(|| {
+// NOTE: This pool is for the main and event workers. The reason why they should
+// separate from the user worker pool is they can starve them if user workers
+// are saturated.
+pub static PRIMARY_WORKER_RT: Lazy<tokio_util::task::LocalPoolHandle> =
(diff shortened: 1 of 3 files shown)


crates/base/src/rt_worker/rt.rs
```diff
@@ -8,7 +8,13 @@ pub static SUPERVISOR_RT: Lazy<tokio::runtime::Runtime> = Lazy::new(|| {
         .unwrap()
 });
 
-pub static WORKER_RT: Lazy<tokio_util::task::LocalPoolHandle> = Lazy::new(|| {
+// NOTE: This pool is for the main and event workers. The reason why they should
+// separate from the user worker pool is they can starve them if user workers
+// are saturated.
+pub static PRIMARY_WORKER_RT: Lazy<tokio_util::task::LocalPoolHandle> =
+    Lazy::new(|| tokio_util::task::LocalPoolHandle::new(2));
```
laktek, line 15: nit: can we extract 2 into a constant? So we can modify it easily or in future can configure.

nyannyacha: Or should we extract it as a constant and expose it as an environment variable like a user worker pool?

nyannyacha: Do you want both?

nyannyacha: I think It seems like a good idea to expose environment variables to quickly prepare for future situations.

laktek: yeah, let's expose it as an environment variable. But let's also set a default of 2 (in case env variable isn't set)

nyannyacha: Done. Please review it again.

PR #249, opened 2024-01-23
feat: allow to ignore `Accept-Encoding` through the CLI argument

nyannyacha:

Feature

## Description

As I've mentioned in the internal chat, this PR is for performance degradation seen in response payloads under `20 bytes`.

In essence, this performance degradation is caused by `streaming compression`, so providing the option to avoid this overhead by adding a CLI argument.

TLDR: This PR provides a CLI option to disable compression by changing the header value of the `Accept-Encoding` of the request to `identity`.

### CLI argument to be created due to the introduction of this PR
* `--ignore-accept-encoding` - Ignore `Accept-Encoding` header in the request


laktek:
What's the use case for this?

nyannyacha:
rebased

PR #250, opened 2024-01-25
Upgrade to deno core 0.243.0

andreespirela:


crates/base/src/deno_runtime.rs
```diff
@@ -435,7 +438,13 @@ impl DenoRuntime {
                 Err(err) => return Poll::Ready(Err(err)),
             };
 
-            let poll_result = js_runtime.poll_event_loop(cx, false);
+            let poll_result = js_runtime.poll_event_loop(
+                cx,
+                PollEventLoopOptions {
+                    wait_for_inspector: false,
+                    pump_v8_message_loop: false,
```
nyannyacha, line 445: 🧐

I guess this option should be set to true. Otherwise, we can't run the tasks V8 has been posted.

Reference: https://groups.google.com/g/v8-users/c/xaSx3ATgFbg/m/8dxCmYZhDgAJ

Cargo.toml
```diff
@@ -58,16 +59,22 @@ regex = "^1.7.0"
...
 urlencoding = { version = "2.1.2" }
-import_map = { version = "0.15.0" }
-base64 = { version = "=0.13.1" }
+import_map = { version = "=0.18.0", features = ["ext"] }
+base64 = "0.21.4"
 futures = { version = "0.3.28" }
 futures-util = { version = "0.3.28" }
 ctor = { version = "0.2.6" }
+percent-encoding = "=2.3.0"
+scopeguard = { version = "1.2.0" }
```
laktek, line 73: do we still using scopeguard cc: @nyannyacha

laktek (approved):

crates/sb_core/net.rs
```diff
@@ -108,7 +108,18 @@ pub async fn op_net_accept(
     if rx.is_none() {
         return Err(bad_resource("unix channel receiver is already used"));
     }
-    let mut rx = rx.unwrap();
+
+    let rx = rx.unwrap();
+    let mut rx = scopeguard::guard(rx, {
+        let state = state.clone();
```
nyannyacha, line 114: This is where `scopeguard` is used.
@laktek

nyannyacha: Since `scopeguard` is being used by two crates(base, sb_core), so i've bumped it up to a workspace dependency.

laktek: ah right! missed it.

PR #254, opened 2024-01-28
fix: `SIGSEGV` has triggered if initializing the main module was failed

nyannyacha:

Bug fix

## Description
If exiting the V8 loop with an exception during script initialization, an isolate was explicitly detached from the thread by `scopeguard`.

It turns out an isolate calls the logic to detach itself from the thread when it is dropped by default, so the introduction of `scopeguard` was wrong.

This PR fixes `SIGSEGV` that triggered due to calling `v8::Isolate::Exit()` twice.

### DIFF
--- a/crates/base/src/deno_runtime.rs
+++ b/crates/base/src/deno_runtime.rs
@@ -345,15 +345,13 @@ impl DenoRuntime {
             op_state.put::<sb_env::EnvVars>(env_vars);
         }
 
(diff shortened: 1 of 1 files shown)


laktek (approved):
Nice work on catching this.

nyannyacha:
It was such a coincidence, I'm glad I found it. 😁

PR #268, opened 2024-02-14
feat: Add experimental Supabase.ai API for running inferences

laktek:

This PR introduces an experimental built-in API to run AI inferences within Edge Functions. This API is currently namespace under `Supabase_UNSTABLE` (check `examples/gte-small-ort` to see how to use it).

To run inferences within Edge Functions, you must [dynamically link Onnx runtime](https://ort.pyke.io/setup/linking). If you use Edge Runtime docker image, this is already configured. But if you wish to compile and run Edge Runtime's AI API locally, you will need to set the `ORT_DYLIB_PATH` environment variable pointing to the local Onnx runtime binary.

Currently, we only support the `GTE` model for generating text embeddings. Hence the model files are included in the Edge Runtime codebase, inside `./models/gte` directory. We'll modify to externally load and configure the models in future releases.


andreespirela (approved):
Lgtm

PR #269, opened 2024-02-15
fix: installation of onnx runtime in x86 plaforms

laktek:

Fix the installation of onnx runtime in x86 platforms.


andreespirela (approved):
Nice script, lgtm

PR #270, opened 2024-02-16
feat: ensure NO_COLOR env variable is repected

krvajal:

Bugfix

console.error is not respecting the NO_COLOR env variable as specified here https://deno.land/api@v1.40.5?s=Deno.noColor. This causes the Log event to contain a msg field which is not plain text but encoded with unicode colors because of https://github.com/denoland/deno/blob/main/ext/console/01_console.js#L2345 

## What is the new behavior?
It respects the NO_COLOR env variable

## Additional context

Add any other context or screenshots.


laktek:
@krvajal Thanks for the PR! Can you make it a `fix:` instead of `feat:` (so we can release it as a patch)?

laktek:
@krvajal Also, it looks like tests & clippy are failing.

krvajal:
@laktek thanks for the review. I will try to address your comments later today.

PR #271, opened 2024-02-18
feat: WebSocket upgrade support

nyannyacha:

Feature

## Description

This PR is to make it possible for the (user/main) worker to upgrade incoming requests to the WebSocket protocol.

### Breaking changes

* `EdgeRuntime.applyConnectionWatcher` is replaced by `EdgeRuntime.applySupabaseTag`
  This change is a result of the need to track multiple pieces of data.

### DIFF
--- a/crates/base/Cargo.toml
+++ b/crates/base/Cargo.toml
@@ -6,6 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+http_utils = { version = "0.1.0", path = "../http_utils" }
 async-trait.workspace = true
 thiserror.workspace = true
 monch.workspace = true
@@ -31,7 +32,7 @@ deno_webidl = { workspace = true }
 deno_web = { workspace = true }
 deno_websocket = { workspace = true }
 httparse = { version = "1.8.0" }
-hyper = { workspace = true, features = ["full"] }
+hyper = { workspace = true, features = ["full", "backports"] }
 http = { version = "0.2" }
 import_map.workspace = true
 log = { workspace = true }
--- a/crates/base/Cargo.toml
+++ b/crates/base/Cargo.toml
@@ -66,7 +66,10 @@ deno_webgpu.workspace = true
 sb_ai = { version = "0.1.0", path = "../sb_ai" }
 
 [dev-dependencies]
+tokio-util = { workspace = true, features = ["rt", "compat"] }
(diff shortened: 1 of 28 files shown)


nyannyacha:
Note: Tests are passed, but there is no test for the upgrade in the WebSocket. I need to write a test.

nyannyacha:
ready for review 😋
cc @andreespirela @laktek

crates/sb_core/js/01_http.js
```diff
@@ -0,0 +1,554 @@
...
+      false,
+    );
+
+    const respondWith = createRespondWith(
+      this,
+      streamRid,
+      abortController,
+    );
+
+    return { request, respondWith, streamRid };
```
nyannyacha, line 174: Note: This is the only purpose this file exists for.
We need `streamRid` for protocol upgrades, but Deno hasn't exposed it at all.

crates/base/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+base_share = { version = "0.1.0", path = "../base_share" }
```
laktek, line 9: Wonder if we should call it `base_share` or can we make it `http_exts` or `http_utils`? Want to keep the crates focused and less confusing. Having a `base_share` crate may cause confusions for other contributors.

nyannyacha: @laktek 
Yeah, `http_utils` would be better.
I’ll reflect that 😁

nyannyacha: done

crates/base/src/rt_worker/worker_ctx.rs
```diff
@@ -83,43 +89,98 @@ async fn handle_request(
...
+
+                    if let Some(mut watcher) = conn_watch {
+                        if watcher.wait_for(|it| *it == ConnSync::Recv).await.is_err() {
+                            error!("cannot track outbound connection correctly");
+                        }
                     }
-                }
 
-                drop(parts);
+                    drop(parts);
```
andreespirela, line 139: Is this necessary?. Does it drop it at the end of the statement anyways?

nyannyacha: I think I just paranoidly put it in there 🙄

nyannyacha: done

laktek (approved):

PR #273, opened 2024-02-19
fix: refactor Supabase AI API

laktek:

Modified to initialize session once and reuse internal session/tokenizer.

### DIFF
--- a/crates/sb_ai/lib.rs
+++ b/crates/sb_ai/lib.rs
@@ -34,6 +34,7 @@ fn init_gte(state: &mut OpState) -> Result<(), Error> {
     let (req_tx, mut req_rx) = mpsc::unbounded_channel::<GteModelRequest>();
     state.put::<mpsc::UnboundedSender<GteModelRequest>>(req_tx);
 
+    #[allow(clippy::let_underscore_future)]
     let _: task::JoinHandle<Result<(), Error>> = task::spawn(async move {
         let session = Session::builder()?
             .with_optimization_level(GraphOptimizationLevel::Disable)?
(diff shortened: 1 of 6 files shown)


crates/sb_ai/lib.rs
```diff
@@ -11,45 +11,52 @@ use tokenizers::Tokenizer;
...
 deno_core::extension!(
     sb_ai,
-    ops = [op_sb_ai_run_model],
+    ops = [op_sb_ai_run_model, op_sb_ai_init_model],
     esm_entry_point = "ext:sb_ai/ai.js",
     esm = ["ai.js",]
 );
 
-fn run_gte(state: &mut OpState, prompt: String) -> Result<Vec<f32>, Error> {
+fn init_gte(state: &mut OpState) -> Result<(), Error> {
```
andreespirela, line 19: It would be interesting if we can make this op `async` and have the whole logic of initialization (if it's heavy) on a different thread. (I'm thinking especially because models can be large)

crates/sb_ai/lib.rs
```diff
@@ -11,45 +11,52 @@ use tokenizers::Tokenizer;
...
-                Path::new(&models_dir)
-                    .join("gte")
-                    .join("gte_small_tokenizer.json"),
-            )
-            .map_err(anyhow::Error::msg)?,
-        )
+        bail!("tokenizer not available. init model first")
     }
-    let mut tokenizer = tokenizer.unwrap();
+    let mut tokenizer = tokenizer.unwrap().clone();
```
andreespirela, line 59: `tokenizer?.clone()` So it the error gets routed to the op

examples/gte-small-ort/index.ts
```diff
@@ -1,7 +1,9 @@
+const model = new Supabase_UNSTABLE.ai.Session('gte-small');
+
 Deno.serve(async (req: Request) => {
 	const params = new URL(req.url).searchParams;
 	const input = params.get('text');
-	const output = Supabase_UNSTABLE.ai.runModel('gte', input);
+	const output = model.run(input);
```
andreespirela, line 6: Also wonder if the `run` op should be async...

crates/sb_ai/lib.rs
laktek: `tokenizer` is an `Option<Tokenizer>` in this case.

andreespirela (approved):
Lgtm!

crates/sb_ai/lib.rs
laktek: @andreespirela Thanks for the suggestion! Refactored the init to run on a tokio task (ie. green thread)

examples/gte-small-ort/index.ts
laktek: Updated

PR #275, opened 2024-02-20
fix: value must be aligned to the time unit being logged

nyannyacha:

Bug fix

## Description

Align the value correctly to the time unit


nyannyacha:
cc @laktek 😋

laktek (approved):

PR #278, opened 2024-02-21
fix: `serve` should handle `onError`

nyannyacha:

Bug fix

## Description

This PR makes `Deno.serve` handles the `onError` option correctly.

```typescript
const onError = (ex) => {
    return new Response(ex.toString(), { status: 200 });
};

Deno.serve({ onError }, () => {
    throw new Error("meow!");
});

// Expected: (200 OK) Error: meow!
```

Reported-by: @divmgl

Fixes supabase/supabase#21395

### DIFF
--- a/crates/sb_core/js/http.js
+++ b/crates/sb_core/js/http.js
@@ -76,17 +76,21 @@ async function serve(args1, args2) {
 
 	if (typeof args1 === 'function') {
 		opts['handler'] = args1;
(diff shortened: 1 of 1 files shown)


divmgl:
Absolutely huge. I would sign off on this but I'm not super familiar with this codebase. Thanks for the quick turnaround

laktek (approved):

PR #281, opened 2024-02-27
feat: TLS listen support

nyannyacha:

Feature

## Description

This PR introduces TLS listening support to edge-runtime.

### List of arguments to be created due to the introduction of this PR

* `--tls [PORT]` - Enables TLS listening. If a port is given after the flag, listen on that port. The default value is `443`. When this flag is given, the user must provide a `--key <Path>` and `--cert <Path>`.
* `--cert <Path>` - Path to PEM-encoded X.509 certificate to be used to TLS. (This includes certificate chains configured as a single file)
* `--key <Path>` - Path to PEM-encoded key to be used to TLS.


crates/base/src/rt_worker/worker_ctx.rs
```diff
@@ -173,12 +174,29 @@ async fn relay_upgraded_request_and_response(
     downstream: OnUpgrade,
     parts: http1::Parts<UnixStream>,
 ) {
+    use std::io::ErrorKind;
```
andreespirela, line 177: Can we move this to the top?

crates/base/src/server.rs
```diff
@@ -188,9 +194,82 @@ pub struct ServerFlags {
...
+            port: self.port,
+            key: self.key.clone_key(),
+            cert_chain: self.cert_chain.clone(),
+        }
+    }
+}
+
+impl Tls {
+    pub fn new(port: u16, key: &[u8], cert: &[u8]) -> anyhow::Result<Self> {
+        use rustls_pemfile::read_one_from_slice;
```
andreespirela, line 216: Can we move these to the top?

crates/base/src/server.rs
```diff
@@ -188,9 +194,82 @@ pub struct ServerFlags {
...
+        use rustls_pemfile::Item;
+
+        let Some((key_item, _)) =
+            read_one_from_slice(key).map_err(|err| anyhow!("can't resolve key: {:?}", err))?
```
andreespirela, line 220: Mhm I'm curious, why `read_one_from_slice` ? The key is fully being pased in the parameter...

crates/base/src/server.rs
```diff
@@ -188,9 +194,82 @@ pub struct ServerFlags {
...
+        else {
+            bail!("invalid key data")
+        };
+
+        let mut cert_chain = vec![];
+        let mut cert_slice = cert;
+        loop {
+            let Some((Item::X509Certificate(cert), remain_cert_slice)) =
+                read_one_from_slice(cert_slice)
```
andreespirela, line 229: Same here, if the `cert_slice` is fully being passed, why do we need to read in loops?

crates/base/src/server.rs
nyannyacha: @andreespirela
I don't understand what you're trying to point out 😅
Yes I am passing the reference to the slice. Was there something better I could have done? 😵‍💫

nyannyacha: Hmmm... This function takes in the entire slice as you said, but returns the slice with the offset added along with the resulting value if it's read successfully.

crates/base/src/server.rs
nyannyacha: Since we may have a certificate file that contains an intermediate certificate, we need to call `read_one_from_slice` to parse the certificate until the `remain_cert_slice` slice tells us that it is empty.

Of course, there are various helper functions in `rustls_pemfile` that can hide loops, but it's a matter of flavor. 😋

crates/base/src/rt_worker/worker_ctx.rs
nyannyacha: solved

crates/base/src/server.rs
nyannyacha: solved

crates/base/src/server.rs
laktek: I think it's better to be explicit. Let's keep it this way.

crates/cli/src/main.rs
```diff
@@ -54,7 +54,27 @@ fn cli() -> Command {
                 .arg(
                     arg!(-p --port <PORT> "Port to listen on")
                         .default_value("9000")
-                        .value_parser(value_parser!(u16)),
+                        .value_parser(value_parser!(u16))
+                )
+                .arg(
+                    arg!(--tls [PORT])
```
laktek, line 60: What happens if a user provides both `--port` and `--tls` port? Can the server listen on both simultaneously?

laktek (approved):

crates/cli/src/main.rs
nyannyacha: > What happens if a user provides both `--port` and `--tls` port? Can the server listen on both simultaneously?

yep, you can usually think of `--port` as being for http and `--tls` as being for https.

nyannyacha: @laktek 

If you specify the arguments as follows. edge-runtime It will listen on two ports for HTTP and TLS (HTTPS), as shown below.

Arguments
```json
[
                "start",
                "-v",
                "--tls",
                "4433",
                "--cert",
                "./localhost+1.pem",
                "--key",
                "./localhost+1-key.pem",
                "-p",
                "9998",
               // ...
            ]
```

Output
```
...
DEBUG FileFetcher::fetch_cached - specifier: https://deno.land/std@0.131.0/async/tee.ts
DEBUG Snapshot already up to date. Skipping pending resolution.
DEBUG Opening cache /home/vscode/.cache/deno/node_analysis_cache_v1...
main function started
DEBUG edge-runtime is listening on 0.0.0.0:9998
DEBUG edge-runtime is listening on 0.0.0.0:4433 (secure)
```

nyannyacha: https://github.com/supabase/edge-runtime/assets/46702285/998d4576-90e9-4238-9871-2da47b366cbb

laktek: Awesome! thanks for taking time to explain how it behaves.

Issue #284, opened 2024-03-08
Documentation: Handle gracefully shutting down server


ajayvignesh01:
# Improve documentation

## Describe the problem

Currently, there is no documentation regarding how to gracefully shut down the server. For example, when using docker compose stop, any pending requests get the ECONNRESET error indicating the server closed its connection. 

I've tried to use Deno.addSignalListener(), but it doesn't seem to be included in the runtime, so lost on how to handle this.

## Describe the improvement

Add some documentation on how to handle gracefully shutting down the server after serving all current requests.

The ideal one would show how to stop new requests, and wait until the current requests finish processing before exiting.

Would also be helpful if there was some documentation on what exactly the global EdgeRuntime does / can do.

nyannyacha:
👀

Ummm....
It is probably not enough to improve the documentation.

It seems like we need to improve the logic that handles signals to consider your case.

nyannyacha:
@ajayvignesh01 I'm working on something around memory metrics right now, and I'll come back to this after I do that 😋

ajayvignesh01:
Thank you @nyannyacha 

Yeah, looking into it more, the enhancement tag does sound right.

I'm mostly needing this for making zero-downtime updates using a simple docker setup for our backend. Currently, we have a pipeline setup that updates the server on a push to main branch, but it comes at the cost of erroring current requests and a few seconds of downtime. In our particular use case, a lot of moving parts are involved that could corrupt data in some places if the function is halted mid response.

On a side note, will open source the repo and make a guide on how to set this up on any VPS when we can get this feat 🫡

nyannyacha:
You're welcome! @ajayvignesh01 😁

Your case is important to me as well.

In particular, edge-runtime currently has a memory leak that is unavoidable without a fork due to an upstream issue, so if we enhance the signal handling to consider this case will allow us to safely restart the container before the OOM-Killer is triggered.

I'm going to make this a high priority, so stay tuned 😋

Have a great day!

ajayvignesh01:
@nyannyacha I wonder if the event-manager in the examples folder could somehow help with this.

Also, here's the repo that I'm trying to get zero-downtime redeploys/upgrades working with: https://github.com/ajayvignesh01/server-tradeloop

Most of the logic for redeploying is in the upgrade.sh script.

nyannyacha:
@ajayvignesh01 

[...]

PR #286, opened 2024-03-15
feat: extra memory regulation layer, revisit termination routines

nyannyacha:

Feature

## Description

This PR introduces an additional layer of tracking and regulating the memory usage used by individual isolates using wakers and timers and garbage collection callback.

V8 uses page allocators to allocate linear memory for web assemblies, so there is no public API in `V8::Isolate` to detect allocations and deallocations.

Furthermore, it turns out that Deno sometimes wraps rust's buffers in `BackingStore` and attaches directly to an isolate, and there is no way for edge-runtime to trap this without a patch.

For the downstream edge-runtime to detect this and regulate excessive memory usage of each isolate, it will need to implement a memory usage tracking routine that is as cost-effective as possible.

This PR is the first plan for that.

### Other things

1. During iterative integration testing, I found an issue with the graceful shutdown routine and it was revisited. 
graceful shutting down routine is related to connection draining (#284) and additional PR will be submitted in the future.

2. Allowed some CLI arguments to be accepted as environment variables. (see [1f32918](https://github.com/supabase/edge-runtime/pull/286/commits/1f32918e0d9d888a4ad049112f53257987c91f74))


crates/base/src/deno_runtime.rs
```diff
@@ -389,12 +468,25 @@ impl DenoRuntime {
             );
         }
 
+        if is_user_worker {
+            js_runtime.v8_isolate().add_gc_prologue_callback(
```
laktek, line 472: @nyannyacha This callback would run before garbage collector would start?

crates/base/src/deno_runtime.rs
```diff
@@ -435,15 +527,44 @@ impl DenoRuntime {
             .load_main_module(&main_module_url, mod_code)
             .await?;
 
+        if is_user_worker {
+            drop(rt::SUPERVISOR_RT.spawn({
+                let drop_token = mem_check_state.drop_token.clone();
+                let waker = mem_check_state.waker.clone();
+
+                async move {
+                    // TODO(Nyannyacha): Should we introduce exponential
```
laktek, line 536: yeah, that'd be neat! Can be a follow up PR?

nyannyacha: @laktek 
Planning to submit a PR in the next iteration.
In my own degradation experiments (see slack), I didn't find any significant degradation, so I just left it as a comment in this PR.

crates/base/src/deno_runtime.rs
nyannyacha: @laktek 
This callback is called just before V8 attempts to garbage collect.[^1][^2]

[^1]:https://source.chromium.org/chromium/chromium/src/+/main:v8/src/heap/heap.cc;l=1774?q=heap.cc
[^2]:https://source.chromium.org/chromium/chromium/src/+/main:v8/src/heap/heap.cc;l=1833?q=heap.cc

laktek (approved):
LGTM! Let's wait for @andreespirela to review it as well.

andreespirela (approved):
Lgtm

nyannyacha:
wait

PR #292, opened 2024-03-21
Configuration options for AI model inference API

laktek:

It's now possible to pass configuration options to AI model's API (currently only supports configuring mean_pool and normalize). 

Also I removed the `UNSTABLE` label from the API.


nyannyacha (approved):
LGTM 😋 (locally tested and it worked normally)

PR #293, opened 2024-03-21
fix: copy models dir from builder

laktek:

#292 broke the build due to a missing directory in docker build. This PR attempts to fix it.


nyannyacha (approved):
LGTM

PR #294, opened 2024-03-21
fix: Fix snapshot logic for fs

andreespirela:


laktek (approved):

crates/cpu_timer/src/lib.rs
```diff
@@ -187,6 +187,7 @@ pub fn get_thread_time() -> Result<i64, Error> {
 
 #[cfg_attr(target_os = "linux", linux::ctor)]
 #[cfg(target_os = "linux")]
+#[allow(static_mut_refs)]
```
nyannyacha, line 190: 👀

nyannyacha: I have identified the context behind this lint and will be submitting a PR to rid.

Ref https://github.com/rust-lang/rust/issues/114447

PR #305, opened 2024-04-02
fix: invalid npm package in statics

andreespirela:


crates/sb_fs/static_fs.rs
```diff
@@ -33,39 +32,9 @@ impl StaticFs {
     }
 
     pub fn is_valid_npm_package(&self, path: &Path) -> bool {
-        if let Some(npm_snapshot) = &self.snapshot {
-            let serialized_snapshot = npm_snapshot.as_serialized();
+        if self.snapshot.is_some() {
             let vfs_path = self.vfs_path.clone();
-            if path.starts_with(vfs_path) {
-                let gen_path_str = path.to_str().unwrap();
```
laktek, line 40: so we don't need the following checks? only checking the vfs_path is enough?

PR #314, opened 2024-04-09
fix: Fix unbundle command to better download function source code

andreespirela:


laktek (approved):

crates/sb_graph/lib.rs
```diff
@@ -199,68 +201,70 @@ pub struct ExtractEszipPayload {
...
     let main_path = PathBuf::from(lowest_path);
     let entry_path = main_path.parent().unwrap();
+
     for global_specifier in specifiers {
-        let module_path = create_module_path(global_specifier, entry_path, output_folder);
+        let full_path = PathBuf::from(global_specifier);
+        let is_insider_path = full_path.starts_with(entry_path);
+
+        let target_folder = if is_insider_path {
+            output_folder.join("src")
```
sweatybridge, line 220: I don't get why it's necessary to create `src` in output directory. Since you've found the lowest common directory, can we just set that as the output directory? Because every path will be children of the lowest common directory.

PR #319, opened 2024-04-15
fix: Handle when Ollama host is not set

laktek:

This PR disables Ollama based inferencing if the host name environment variable is not set.

### DIFF
--- a/crates/sb_ai/ai.js
+++ b/crates/sb_ai/ai.js
@@ -40,13 +40,15 @@ const parseJSON = async function* (itr) {
 class Session {
 	model;
 	is_ollama;
+	inferenceAPIHost;
 
 	constructor(model) {
 		this.model = model;
 		this.is_ollama = false;
 
+		this.inferenceAPIHost = core.ops.op_get_env('AI_INFERENCE_API_HOST');
 		if (model === 'llama2' || model === 'mixtral' || model === 'mistral') {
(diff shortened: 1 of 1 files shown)


crates/sb_ai/ai.js
```diff
@@ -40,29 +40,30 @@ const parseJSON = async function* (itr) {
...
+	inferenceAPIHost;
 
 	constructor(model) {
 		this.model = model;
 		this.is_ollama = false;
 
+		this.inferenceAPIHost = core.ops.op_get_env('AI_INFERENCE_API_HOST');
 		if (model === 'llama2' || model === 'mixtral' || model === 'mistral') {
-			this.is_ollama = true;
+			this.is_ollama = !!this.inferenceAPIHost; // only enable ollama if env variable is set
```
nyannyacha, line 51: I looked at the PR the other day and it looks like it's taking a URL format.
Rather than just checking that the variable is declared, why not parse it with something like [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL) to be sure?

nyannyacha (approved):

crates/sb_ai/ai.js
laktek: I think checking the variable existence is good enough. Trying to parse could be slower and error-prone.

PR #321, opened 2024-04-16
perf: optimise hash function to use single memory allocation

AayushyaVajpayee:

Performance optimisation

Currently the gen function performs atleast 33 separate string memory allocations which  can be avoided
## What is the new behavior?
Init a string with 64 bytes and push hashed bytes into that. This is only 1 allocation
Feel free to include screenshots if it includes visual changes.

## Additional context

Add any other context or screenshots.

### DIFF
--- a/crates/sb_core/util/checksum.rs
+++ b/crates/sb_core/util/checksum.rs
@@ -18,6 +18,9 @@ mod tests {
     fn test() {
         let input = vec![b"hello", b"world", b"hello", b"hello"];
         let output = gen(&input);
-        assert_eq!("00d03979f1c2c9cece94003e15ced43d1de8bf126f28d27b93f1e37874fb5395",output.as_str());
+        assert_eq!(
+            "00d03979f1c2c9cece94003e15ced43d1de8bf126f28d27b93f1e37874fb5395",
+            output.as_str()
+        );
     }
 }
--- a/crates/sb_core/util/checksum.rs
+++ b/crates/sb_core/util/checksum.rs
@@ -2,22 +2,12 @@
 
 use ring::digest::Context;
 use ring::digest::SHA256;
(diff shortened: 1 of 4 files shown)


crates/sb_core/util/checksum.rs
```diff
@@ -2,17 +2,32 @@
...
-    let out: Vec<String> = digest
+    let mut hash_str = String::with_capacity(64);
+    digest
         .as_ref()
         .iter()
-        .map(|byte| format!("{byte:02x}"))
-        .collect();
-    out.join("")
+        .for_each(|byte| write!(hash_str, "{byte:02x}")
+            .expect("write! macro on string cannot fail"));
```
andreespirela, line 18: Let’s use faster_hex similar to the current Deno https://github.com/denoland/deno/blob/main/cli/util/checksum.rs

AayushyaVajpayee: faster_hex is fast!!. did a little benchmarking, for the three approaches of gen function.

1. using existing string concatenation
2. using write! macro
3. using faster_hex. 
here are the results:

|           | string concate | write! macro | faster-hex |
|-----------|----------------|--------------|------------|
| mean time | 2.791 us          | 973 ns        | 260 ns      |   

also attaching the violin plot for results

Screenshot 2024-04-17 at 6 44 18 PM

andreespirela: Amazing thanks!

andreespirela (approved):
lgtm! Thanks for the contribution!

AayushyaVajpayee:
@andreespirela I have fixed the formatting related faliure, can you approve workflow run please

andreespirela:
@AayushyaVajpayee thank you for your contribution!

PR #320, opened 2024-04-16
feat: use `DuplexStream` instead of `UnixStream` to communicate with workers

nyannyacha:
## What kind of change does this PR introduce?

Refactor

## Description

This PR changes the stream type used to communicate between workers to `DuplexStream` instead of `UnixStream` to improve request throughput.

It was tested on a local machine, so it's imprecise, but it showed an improvement of about 10% in request throughput per second.

### List of arguments to be created due to the introduction of this PR
* `--tcp-nodelay` flag disables Nagle's algorithm. (Default is `true`)

Figure1. UnixStream, Debug
unixstream_debug

Figure2. DuplexStream, Debug
duplexstream_debug

Figure3. UnixStream, Release
unixstream_release

Figure4. DuplexStream, Release
duplexstream_release

Figure5. DuplexStream, Debug (examples/chunked-text, TCP_NODELAY OFF)
duplex-chunked-debug

Figure6. DuplexStream, Debug (examples/chunked-text, TCP_NODELAY ON)
스크린샷 2024-04-17 오후 1 32 45

PR #327, opened 2024-04-23
feat: reintroduce `DuplexStream`

nyannyacha:

Bug fix, Enhancement

## Description

> This PR is a continuation of [PR-320](https://github.com/supabase/edge-runtime/pull/320). Please see PR-320 for background on the PR.

### Makes the timing of pipe closure is adjustable within a separately created tokio task, so that the stream is not closed in the middle of a response.

In the case of `DuplexStream`, if either pair is dropped, the stream immediately goes into an unwritable state.

These streams must remain writable for outside responses, but the previous PR did not. So this issue was immediately exposed by the flush behavior of libraries like `hyper`.

With the current implementation, which requires a Tokio task to be created every time a stream is dropped, I get a small performance hit compared to PR-320, but I believe we can replace it with an efficient implementation in the next iteration.

### `ConnSync` has been removed and replaced with `CancellationToken`.

This was a refactoring to simplify the structure, and while the performance impact of the replacement has not been carefully evaluated, it does not appear to have changed much.

### Stream of request body must be cancelable from errors or exceptions by the worker.

Sometimes the request body is not fully consumed by the worker due to circumstances such as worker resource limit due to a multipart request. 

Previously, `sb_workers` did not handle these circumstances properly, leaving the end requestor in a situation where they would receive no response indefinitely. (Runtime also had resource leaks due to unresolved promises)

This PR introduces a routine that tries to cancel the request body stream after checking for a response or exception from the worker.

### Graceful shutdown by `SIGWINCH` is now behind the `base/termination-signal-ext` feature flag.

Given the nature of devs frequently changing the terminal size, SIGWINCH has been gated to only be enabled behind a specific feature flag.


scripts/run.sh
```diff
@@ -1,3 +1,6 @@
 #!/usr/bin/env bash
 
-GIT_V_TAG=0.1.1 cargo build && EDGE_RUNTIME_PORT=9998 RUST_BACKTRACE=full ./target/debug/edge-runtime "$@" start --main-service ./examples/main --event-worker ./examples/event-manager
+GIT_V_TAG=0.1.1 cargo build && EDGE_RUNTIME_PORT=9998 RUST_BACKTRACE=full ./target/debug/edge-runtime "$@" start \
+    --main-service ./examples/main \
+    --event-worker ./examples/event-manager \
+    --request-wait-timeout 200000
```
laktek, line 6: what's the `request-wait-timeout` flag?

laktek (approved):

PR #330, opened 2024-04-29
fix: do not shutdown server on resize

laktek:

do not process `SignalKind::window_change()]`


nyannyacha:
Are you making pr because of DX?

PR #333, opened 2024-05-02
fix: make the wall clock limit handle some edge cases correctly

### LINKED ISSUE #99: runtime closes connection when worker exceeds timeout
# Bug report

I'm running using the docker image v1.3.3 on a Mac M1.
When a worker is taking too long the runtime errors and stops working.  

## Describe the bug

The worker has a timeout of 100 milliseconds.  
The hello-world worker is waiting for 1 second before returning a response.  

The error is:  
```
web_1  | main function started
web_1  | serving the request with /home/deno/functions/hello-world
web_1  | hello-world started
web_1  | Error: channel closed
web_1  |     at async UserWorker.fetch (ext:sb_user_workers/user_workers.js:54:21)
web_1  |     at async Server.<anonymous> (file:///home/deno/functions/main/index.ts:39:16)
web_1  |     at async Server.#respond (https://deno.land/std@0.131.0/http/server.ts:219:24)
web_1  | serving the request with /home/deno/functions/hello-world
web_1  | Error: channel closed
web_1  |     at async Function.create (ext:sb_user_workers/user_workers.js:80:21)
web_1  |     at async Server.<anonymous> (file:///home/deno/functions/main/index.ts:31:24)
web_1  |     at async Server.#respond (https://deno.land/std@0.131.0/http/server.ts:219:24)
```

The first request throws in `ext:sb_user_workers/user_workers.js:54:21`.   
Every subsequent request throws in `ext:sb_user_workers/user_workers.js:80:21`.  

## To Reproduce

I created a fork from the self hosted example.
https://github.com/EzzatOmar/self-hosted-edge-functions-demo

## System information

- OS: macOs
- Docker image: ghcr.io/supabase/edge-runtime:v1.3.5

nyannyacha:

Bug fix

## Description

```typescript
// Inputs below will disable the wall clock limit.
const workerTimeoutMs = -999;
const workerTimeoutMs = -1;
const workerTimeoutMs = 0;
const worker = EdgeRuntime.userWorkers.create({
    ...,
    workerTimeoutMs
});
```

```typescript
// Worker is created and immediately terminated by the wall clock limit. However, if the script
// is lightweight enough to initialize quickly, the request can be processed before the worker exits.
const workerTimeoutMs = 1;
const worker = EdgeRuntime.userWorkers.create({
    ...,
    workerTimeoutMs
});
```

```typescript
// main.ts
const workerTimeoutMs = 2000;
const worker = EdgeRuntime.userWorkers.create({
    ...,
    workerTimeoutMs
});

// serve.ts
async function sleep(ms: number) {
    return new Promise(res => {
        setTimeout(() => {
            res(void 0);
        }, ms)
    });
}

// Wait 3000 ms on purpose. (Top-level await)
await sleep(3000);
Deno.serve((_req) => new Response("Hello, world"));

// Following case is also valid.
// Deno.serve(async () => {
//     await sleep(3000);
//     return new Response("Hello, world");
// });

// This script will be forced to exit after about 2000 ms with no requests processed.
```

Fixes #99

### DIFF
--- a/crates/base/src/rt_worker/implementation/default_handler.rs
+++ b/crates/base/src/rt_worker/implementation/default_handler.rs
@@ -2,9 +2,12 @@ use crate::deno_runtime::DenoRuntime;
 use crate::rt_worker::supervisor::CPUUsageMetrics;
 use crate::rt_worker::worker::{HandleCreationType, UnixStreamEntry, Worker, WorkerHandler};
 use anyhow::Error;
-use event_worker::events::{BootFailureEvent, PseudoEvent, UncaughtExceptionEvent, WorkerEvents};
+use event_worker::events::{
+    BootFailureEvent, EventLoopCompletedEvent, UncaughtExceptionEvent, WorkerEvents,
+};
 use log::error;
 use std::any::Any;
+use std::time::Duration;
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender};
 use tokio::sync::oneshot::Receiver;
 
@@ -50,7 +53,23 @@ impl WorkerHandler for Worker {
                         }))
                     }
                 }
-                (Ok(()), _) => Ok(WorkerEvents::EventLoopCompleted(PseudoEvent {})),
+
+                (Ok(()), cpu_usage_ms) => {
+                    if created_rt.is_termination_requested.is_raised() {
+                        static EVENT_RECV_DEADLINE_DUR: Duration = Duration::from_secs(5);
+
+                        if let Ok(Ok(ev)) =
+                            tokio::time::timeout(EVENT_RECV_DEADLINE_DUR, termination_event_rx)
+                                .await
+                        {
+                            return Ok(ev.with_cpu_time_used(cpu_usage_ms as usize));
+                        }
+                    }
+
+                    Ok(WorkerEvents::EventLoopCompleted(EventLoopCompletedEvent {
+                        cpu_time_used: cpu_usage_ms as usize,
+                    }))
+                }
             }
         };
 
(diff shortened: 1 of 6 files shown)


crates/base/src/rt_worker/implementation/default_handler.rs
```diff
@@ -50,7 +53,23 @@ impl WorkerHandler for Worker {
                         }))
                     }
                 }
-                (Ok(()), _) => Ok(WorkerEvents::EventLoopCompleted(PseudoEvent {})),
+
+                (Ok(()), cpu_usage_ms) => {
+                    if created_rt.is_termination_requested.is_raised() {
+                        static EVENT_RECV_DEADLINE_DUR: Duration = Duration::from_secs(5);
```
laktek, line 59: why do we need to wait for 5 seconds here?

It's nice we can account for CPU time usage on event loop completed events too.

laktek (approved):

crates/base/src/rt_worker/implementation/default_handler.rs
nyannyacha: @laktek It is an arbitrarily chosen time in case it might not be resolved forever.

nyannyacha: Note:
I'm revisiting this part, and there's a possibility that if a supervisor unconditionally requests the isolate to terminate, it might not come to this branch, so it might be removed in the future.

laktek: In that case can we add a comment on it?

nyannyacha: @laktek 
Yes, I will add that.

PR #338, opened 2024-05-10
feat: expose systemMemoryInfo API for main worker

laktek:
Adding it to the `EdgeRuntime` namespace instead of the `Deno` namespace to prevent accidentally exposing it in user workers. Also, we are implementing a customized version of it for performance reasons (ie. only support in Linux via sysinfo)


nyannyacha (approved):
LGTM 😋

PR #343, opened 2024-05-15
perf: implement lazy load for eszip entities

nyannyacha:

Enhancement

## Description

This PR changes the way opaque data is stored in eszip and makes allow lazy-load for individual data stored in eszip.

### Sample
```typescript
import { createClient } from 'npm:@supabase/supabase-js@2.42.0'
console.log(typeof createClient);
Deno.serve((_req) => new Response("Hello, world"));
```

### Metrics (Observed lowest value in local)

||Previous|This PR|Change (%)|
|-|---------------------|------------------|-----------------|
|Bundle Size|22MiB|6.3MiB|-349.2%|
|Boot Time|1939ms|84ms|-2308.3%|

### Notes
It supports runtime migration for bundles created before this PR. However, performance will revert back to what it was before, so I recommend regenerating the bundle.


nyannyacha:
Note:

Currently, a single entity in eszip contains a big opaque data for virtual fs.

I think we need to bring it up to the eszip entity level to improve performance.

nyannyacha:
Note:
Once this PR is merged, it is worth looking closely to ensure that it is completely compatible with the previous eszip implementation.

cc @laktek @andreespirela

crates/sb_fs/file_system.rs
```diff
@@ -33,22 +35,25 @@ impl DenoCompileFileSystem {
         }
     }
 
-    fn copy_to_real_path(&self, oldpath: &Path, newpath: &Path) -> FsResult<()> {
```
andreespirela, line 36: why removing this function? We could just create an async one?

nyannyacha: I recall that some of the functions it calls in its function body had their signatures changed to async, so I changed those functions to async due to contagious.

nyannyacha: I guess I could probably make that synchronously again by spawning async functions in the blocking threads

laktek (approved):
LGTM! Did some testing locally and common scenarios seems to work :)

PR #350, opened 2024-05-27
fix: don't emit `RequestCancelledBySupervisor` when event loop error has been raised

nyannyacha:

Enhancement

## Description

This PR makes to record `InvalidWorkerResponse` with an event loop error message in the response instead of emitting `RequestCancelledBySupervisor` if an uncaught error is encountered by the user script.

> Note: Error message recorded in the response may be something other than an event loop error message due to race conditions caused by a worker exiting quickly. (but error kind is still fixed to `InvalidWorkerResponse`)

Example
```typescript
// Throw an exception from the top level (will cause an event loop error and a worker will be terminated immediately)
throw new Error("meow");

// Deno.serve((_req) => new Response("meow"));
```

Response (Before this PR)
```jsonc
// 500 Internal Server Error
{
    "msg": "WorkerRequestCancelled: request has been cancelled by supervisor"
}
```

Response (This PR)
```jsonc
// 500 Internal Server Error
{
    "msg": "InvalidWorkerResponse: event loop error: Error: meow!\n    at file:///<location>"
}
```

### DIFF
--- a/crates/base/src/rt_worker/worker.rs
+++ b/crates/base/src/rt_worker/worker.rs
@@ -12,7 +12,7 @@ use event_worker::events::{
 use futures_util::FutureExt;
 use log::{debug, error};
 use sb_core::{MetricSource, RuntimeMetricSource, WorkerMetricSource};
-use sb_workers::context::{UserWorkerMsgs, WorkerContextInitOpts};
+use sb_workers::context::{UserWorkerMsgs, WorkerContextInitOpts, WorkerExit, WorkerExitStatus};
 use std::any::Any;
 use std::future::{pending, Future};
 use std::pin::Pin;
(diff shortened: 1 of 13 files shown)


crates/base/src/rt_worker/worker.rs
```diff
@@ -269,13 +270,19 @@ impl Worker {
...
 
-                            let found_unexpected_error = result.is_err()
-                                || matches!(
-                                    result.as_ref(),
-                                    Ok(WorkerEvents::UncaughtException(_))
-                                );
+                            let maybe_uncaught_exception_event = match result.as_ref() {
+                                Ok(WorkerEvents::UncaughtException(ev)) => Some(ev.clone()),
+                                Err(err) => Some(UncaughtExceptionEvent {
+                                    cpu_time_used: 0,
```
laktek, line 276: I guess this is here as a placeholder value and gets replaced later?

laktek (approved):

crates/base/src/rt_worker/worker.rs
nyannyacha: @laktek 
That arm is the result of the return type signature of `handle_creation`.
In fact, that arm is unreachable.

As you can see in `default_handler.rs`, nowhere in the body of the `handle_creation` does it return an Err.

laktek: ok, gotcha

PR #352, opened 2024-05-30
Integrate deno 1 43 0

andreespirela:


crates/node/ops/os/mod.rs
```diff
@@ -72,19 +72,6 @@ where
...
-{
-    {
-        let permissions = state.borrow_mut::<P>();
-        permissions.check_sys("abort", "node:process.abort()")?;
-    }
-
-    std::process::abort();
-}
-
 #[op2]
```
nyannyacha, line 75: It seems to have been reverted.

crates/node/ops/process.rs
```diff
@@ -0,0 +1,64 @@
...
+    state
+        .borrow_mut::<PermissionsContainer>()
+        .check_run_all("process.kill")?;
+    Ok(kill(pid, sig))
+}
+
+#[op2(fast)]
+pub fn op_process_abort() {
+    std::process::abort();
+}
```
nyannyacha, line 64: It seems to have been reverted.

crates/node/polyfills/http.ts
```diff
@@ -1640,28 +1677,14 @@ export class ServerImpl extends EventEmitter {
...
         });
-
-        tag.fenceRid = fenceRid;
+        // Update socket held by `req`.
+        req.socket = socket;
         this.emit("upgrade", req, socket, Buffer.from([]));
-
-        return internals.RAW_UPGRADE_RESPONSE_SENTINEL;
+        return response;
       } else {
```
nyannyacha, line 1688: This seems to have been reverted as well.

crates/sb_os/lib.rs
```diff
@@ -18,7 +18,7 @@ pub struct MemInfo {
 #[op2]
 #[serde]
 fn op_system_memory_info() -> Option<MemInfo> {
-    let mut mem_info = MemInfo {
+    let mem_info = MemInfo {
         total: 0,
```
nyannyacha, line 22:
```suggestion
    let mut mem_info = MemInfo {
        total: 0,
```

crates/node/ops/fs.rs
```diff
@@ -78,3 +79,146 @@ where
...
+    pub blocks: u64,
+    pub bfree: u64,
+    pub bavail: u64,
+    pub files: u64,
+    pub ffree: u64,
+}
+
+#[op2]
+#[serde]
+pub fn op_node_statfs<P>(
```
laktek, line 97: I guess this should be a no-op? We don't expose Deno's FS API too.

crates/node/ops/mod.rs
```diff
@@ -7,8 +7,12 @@ pub mod http2;
 pub mod idna;
 pub mod ipc;
 pub mod os;
+pub mod process;
 pub mod require;
 pub mod util;
 pub mod v8;
+pub mod vm;
```
laktek, line 14: do we need to expose `vm`? Probably need to stay disabled.

crates/node/ops/mod.rs
```diff
@@ -7,8 +7,12 @@ pub mod http2;
...
+mod vm_internal;
 pub mod winerror;
+pub mod worker_threads;
```
laktek, line 17: This should probably staty disabled as well.

crates/node/ops/os/mod.rs
laktek: @nyannyacha btw, we shouldn't expose `process.abort()` right? or API can be there but it should be a no-op.

crates/node/ops/process.rs
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+#[op2(fast)]
+pub fn op_node_process_kill(
+    state: &mut OpState,
+    #[smi] pid: i32,
+    #[smi] sig: i32,
+) -> Result<i32, AnyError> {
+    state
```
laktek, line 55: This should be a no-op.

crates/node/ops/vm.rs
```diff
@@ -0,0 +1,127 @@
...
+        let result = self
+            .inner
+            .eval_machine(&mut scope, context)
+            .unwrap_or_else(|| v8::undefined(&mut scope).into());
+        Ok(scope.escape(result))
+    }
+}
+
+#[op2]
+pub fn op_vm_create_script<'a>(
```
laktek, line 60: All of vm ops should be no-ops.

crates/node/ops/worker_threads.rs
```diff
@@ -0,0 +1,83 @@
...
+    P: NodePermissions + 'static,
+{
+    let resolver = state.borrow::<NpmResolverRc>();
+    let permissions = state.borrow::<P>();
+    resolver.ensure_read_permission(permissions, file_path)
+}
+
+#[op2]
+#[string]
+pub fn op_worker_threads_filename<P>(
```
laktek, line 29: Looks like worker threads are disabled. Do we need to add this op?

andreespirela: We could probably remove this file

crates/node/ops/vm.rs
andreespirela: Sure

crates/node/ops/process.rs
andreespirela: Sure

crates/node/ops/mod.rs
andreespirela: Perfect

crates/node/ops/mod.rs
andreespirela: Sure

crates/node/ops/fs.rs
andreespirela: Sure

crates/node/ops/os/mod.rs
nyannyacha: @laktek 

Exposing this function to a user worker without permission checks leaves open the possibility that malicious code could abort the process.

Making this function to no-op as you say is one way to go.

crates/node/polyfills/worker_threads.ts
```diff
@@ -1,20 +1,53 @@
...
-import { isAbsolute, resolve } from "node:path";
+import { core, internals, primordials } from "ext:core/mod.js";
+import {
+  op_create_worker,
+  op_host_post_message,
+  op_host_recv_ctrl,
+  op_host_recv_message,
+  op_host_terminate_worker,
+  op_message_port_recv_message_sync,
+  op_worker_threads_filename,
```
nyannyacha, line 12: You removed `op_worker_threads_filename` in rust land, but in js land, call site still exists. Wouldn't this be a problem?
I think there are a few more cases of this.

nyannyacha:
[Blacklisted API](https://github.com/supabase/edge-runtime/blob/0b666d037a3ddffeaac6dca47ffa169a17df287c/crates/base/test_cases/user-worker-san-check/.blacklisted)

Garett-MacGowan:
ETA on this?

crates/node/polyfills/os.ts
```diff
@@ -338,37 +339,37 @@ export function uptime(): number {
...
+  //   _homedir = _homedir ? Buffer.from(_homedir) : _homedir;
+  //   shell = shell ? Buffer.from(shell) : shell;
+  //   username = Buffer.from(username);
+  // }
 
   return {
-    uid,
-    gid,
+    uid: 1000,
+    gid: 1000,
```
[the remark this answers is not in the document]
andreespirela: Good one

crates/node/polyfills/os.ts
```diff
@@ -188,7 +197,8 @@ export function homedir(): string | null {
       return Deno.env.get("HOME") || null;
     default:
       throw Error("unreachable");
-  }
+  }*/
+  return "/src"
```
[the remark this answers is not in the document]
andreespirela: Did /src bc of Deno deploy (it's what Deno deploy returns)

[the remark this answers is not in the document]
laktek: hmm..let's go with `/home/deno` (that's what eszip bundle use as the entrypoint path)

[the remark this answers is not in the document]
nyannyacha: If that's the return value of deno deploy, I think we're good to go.

nyannyacha (approved):
LGTM (Thanks for bringing this!)

PR #361, opened 2024-06-12
fix: bump rust builder image to `1.78-bookworm`

nyannyacha:

Bump rust builder image to `1.78-bookworm`


nyannyacha:
cc @laktek

laktek (approved):

PR #374, opened 2024-06-21
fix(sb_workers): don't propagate promise rejection on request body streaming failure

### LINKED ISSUE #366: TypeError request body receiver not connected (channel closed)
This seems to happen at Function Boot for some functions.

Stack trace:

TypeError
request body receiver not connected (channel closed)

ext:deno_web/06_streams.js in Object.write at line 1155:20
ext:deno_webidl/00_webidl.js in Module.invokeCallbackFunction at line 944:16
ext:deno_web/06_streams.js in WritableStreamDefaultController.writeAlgorithm at line 3915:14
ext:deno_web/06_streams.js in writableStreamDefaultControllerProcessWrite at line 4485:55
ext:deno_web/06_streams.js in writableStreamDefaultControllerAdvanceQueueIfNeeded at line 4388:5
ext:deno_web/06_streams.js in writableStreamDefaultControllerWrite at line 4532:3
ext:deno_web/06_streams.js in writableStreamDefaultWriterWrite at line 4679:3
ext:deno_web/06_streams.js in Object.chunkSteps at line 2743:17
ext:deno_web/06_streams.js in readableStreamFulfillReadRequest at line 2585:17
ext:deno_web/06_streams.js in readableByteStreamControllerCommitPullIntoDescriptor at line 2276:5

nyannyacha:

Bug fix

## Description

For a promise to a WritableStream that streams the body of an inbound request to a user worker, the promise might be rejected in some cases.

Previously, if this promise was rejected, an exception to this rejection would be propagated to the caller.

```typescript
let [sent, res] = await Promise.allSettled([reqBodyPromise, resPromise]);
```

> The promise `res`, which gets the response from the user worker, and the promise `sent`, which streams the request body to the user worker, are bound by `Promise.allSettled`, so it will not return until all of the passed Promises have performed their actions.

However, propagating exceptions for these `sent` is turns out to be undesirable. 

This is because the user worker may consume the request object abnormally, or `op_user_worker_fetch_send` may drop the channel object in the body under the request object, causing the promise to be rejected.

Therefore, we should always expose only the result for `res` to the caller.

Fixes #366

### DIFF
--- a/crates/base/tests/integration_tests.rs
+++ b/crates/base/tests/integration_tests.rs
@@ -714,6 +714,7 @@ async fn test_file_upload_real_multipart_bytes() {
     test_oak_file_upload(
         Cow::Borrowed("./test_cases/main"),
         (9.98 * MB as f32) as usize, // < 10MB (in binary)
+        None,
         |resp| async {
             let res = resp.unwrap();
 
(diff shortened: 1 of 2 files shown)


laktek (approved):
LGTM! should we add a test for this?

PR #377, opened 2024-06-24
fix: mitigate denoland/deno_core#762

nyannyacha:

Bug fix

## Description

There are cases where the specifier is mapped to a different alias by the module loader.

In this case, the waker for the future that resolves the module in the upstream codebase may not be properly registered, causing the module resolution phase to hang indefinitely.

This upstream bug has been fixed in denoland/deno_core#762, but since we don't plan to version up the Deno codebase yet, we decided to fork and cherry-pick that PR.


nyannyacha:
cc @laktek 

https://github.com/supabase/edge-runtime/compare/main...nyannyacha:fix-concurrent-import-no-waker?expand=1#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R84-R86

Cargo.toml
```diff
@@ -81,9 +81,9 @@ http = "0.2"
 faster-hex = "0.9.0"
 tracing = "0.1"
 
-# DEBUG
-# [patch.crates-io]
-# deno_core = { path = "/Users/andrespirela/Documents/workspace/supabase/deno_core/core" }
+[patch.crates-io]
+# TODO(Nyannyacha): Patch below is temporary. Clean the line in the Deno 1.44 update.
+deno_core = { git = "https://github.com/Nyannyacha/deno_core", branch = "278-supabase" }
```
laktek, line 86: use the supabase clone https://github.com/supabase/deno_core/tree/278-supabase

laktek (approved):

PR #386, opened 2024-07-10
feat: support setting network permissions via allowNet option

0x5457:

feature

## What is the new behavior?

Allow network access.  can specify an optional list of IP addresses or hostnames (optionally with ports) to provide an allow-list of allowed network addresses.

```ts
EdgeRuntime.userWorkers.create({
...
  allowNet: ["<IP/HOSTNAME>"]
...
})

```


Code2Life:
this feature is very useful for fine-grained networking control👍

Codexiaoyi:
also need this feature to me!

laktek:
Can you explain the use cases where you need explicit network access? I guess this would be fine to add for self-hosting, but unlikely we'd provide an option to configure it in hosted environment.

0x5457:
yes, this is useful for self-hosting cases, especially when EdgeRuntime used standalone rather than with Supabase.

It could extends the EdgeRuntime use case, for example, when the user workers need to be NOT homogeneous, fetching some metadata from localhost is useful, and keep the other network access blocked to avoid security risk.

Another reason is Deno has this option, it would be great if EdgeRuntime could align with Deno
https://docs.deno.com/runtime/manual/basics/permissions/#permissions-list

nyannyacha (approved):
LGTM!
Thanks for bringing this feature!

cc @laktek

laktek:
Yeah, LGTM too! Thanks @0x5457 for the contribution and @nyannyacha for adding tests.

(This won't be configurable via Supabase CLI or on hosted platform for now. We'll do a separate Changelog announcement when we enable it)

krvajal:
could be possible to add the a deny list as well?

PR #404, opened 2024-09-03
fix: inspector causes panic

nyannyacha:

Bug fix

## Description

It seems that the path to calling `thread::unpark` was enabled when eszip content became possible to load asynchronously.

The current inspector implementation upstream doesn't like it when `park` is called at a certain stage of `poll_session` and then immediately unparked. This results in reaching `unreachable!`, which causes panic.

### Related Links
* https://github.com/supabase/cli/issues/2646
* https://github.com/supabase/deno_core/commit/d73dc1e1845d00cc2e8d2a32caf5859cc300386b

Reported-by: @daniel-sc

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1301,8 +1301,7 @@ dependencies = [
 [[package]]
 name = "deno_core"
 version = "0.293.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


laktek:
Squash and merge please

laktek (approved):

PR #368, opened 2024-06-18
feat: add support for custom ai models

kallebysantos:
>**Since @nyannyacha had made a code refactor, maybe the original description does not reflect the architecture changes anymore, but you can still read it bellow:**

<details>
  <summary>Original description</summary>

## What kind of change does this PR introduce?

Feature, Enhancement

## What is the current behavior?

As described in the [following discussion](https://github.com/orgs/supabase/discussions/23632), currently `edge-runtime` only allows the coupled `gte-small` as `feature-extraction` model.

## What is the new behavior?

This PR introduces a new Rust definition of a `transformers` like API to `edge-runtime`. So it allows to install different `ONNX` models at same time, as well contains the base API definitions to support other kinds of `pipelines` rather than just `feature-extraction` task. 

This new API introduces the `Pipeline` class, that pretends to be similar as [`xenova/transformers`](https://github.com/xenova/transformers.js) but backed by Rust. The `Pipeline` class aim to be an evolution of the current `Session` class, where only support `gte-small` for feature extractions and `ollama` models. The new class defines APIs that allows:
- Different kinds of `tasks`, `feature-extraction`(implemented), `token-classification`, `sentiment-analysis` and others ...
- Usage of pre-installed models
- GPU support provided by [`ort`](https://ort.pyke.io/)
> GPU support was added in the [ort-gpu-runtime](https://github.com/kallebysantos/edge-runtime/tree/ort-gpu-runtime) branch, see also [ort - Execution Providers](https://ort.pyke.io/perf/execution-providers#registering-execution-providers)

### Docker Image:
You can get a docker image of this from docker hub:
```bash
# default runtime
docker pull kallebysantos/edge-runtime:latest

# gpu with cuda provider
docker pull kallebysantos/edge-runtime:latest-cuda
```

### Pipeline usage:
To use the new `Pipeline` class is very simple and looks very similar to `Session`:

**Using default `gte-small`, pre-installed by `Supabase` team:**
```typescript
// By default `feature-extraction` will map to `gte-small`
const pipe = new Supabase.ai.Pipeline('feature-extraction');

Deno.serve(async (req: Request) => {
        const embeddings = await pipe.run("Hello World", { mean_pool: true, normalize: true });
	
	return new Response(JSON.stringify(embeddings), { status: 200 });
});
```

**Using different model:**
```typescript
// Just set the model name, similar to `xenova/transformers` or python `transformers`
const pipe = new Supabase.ai.Pipeline(
	'feature-extraction',
[...]

PR #430, opened 2024-10-28
feat: onnx runtime shared sessions 

kallebysantos:
> This PR is an adapted part of #368 work, that was closed due proposal changing

Feature, Enhancement

model sessions are eager evaluated and not survive cross worker life-cycles

## What is the new behavior?

image

This PR introduces **shared sessions** logic and more other **ort** improvements like GPU support and optimizations. 

> 👷 It's also a base foundation for an owned **onnx** runtime that can be integrated direclty with [huggingface/transformers.js](https://github.com/huggingface/transformers.js/pull/947) library, that will allow a better inference without the needed of coupling models like `gte-small` on `edge-runtime` image.

## Tester docker image:

You can get a docker image of this PR from docker hub:
```bash
# default runtime
docker pull kallebysantos/edge-runtime:latest

# gpu with cuda provider
docker pull kallebysantos/edge-runtime:latest-cuda
```

### Session lifecycle:
This PR introduces a `Lazy` map of `ort:Sessions`, it means that sessions will be loaded once and then shared between worker cycles. 

**Cleaning up sessions:**
Each `ort:Session` is attached to an `Arc` smart pointer and will only be dropped if no consumer is attached to it, but in order to that users must explicit call the `EdgeRuntime.ai.tryCleanupUnusedSession()` method.
> **NOTE:** This method is only available for the `main worker`

```typescript
// cleanup unused sessions every 30s
setInterval(async () => {
  const { activeUserWorkersCount } = await EdgeRuntime.getRuntimeMetrics();
  if (activeUserWorkersCount > 0) {
    return;
  }
  try {
    const cleanupCount = await EdgeRuntime.ai.tryCleanupUnusedSession();
    if (cleanupCount == 0) {
      return;
    }
    console.log('EdgeRuntime.ai.tryCleanupUnusedSession', cleanupCount);
  } catch (e) {
    console.error(e.toString());
  }
}, 30 * 1000);
```

### GPU Support:
The `gpu` support allows `session` inference in specialized hardware and its backed with CUDA. There is no configuration to do by the final user, just call the `Session`  for `gte-small`. But in order to enable `gpu` inference the `Dockerfile` now has two main `build stages` *(That should be specified during `docker build`)* :

**edge-runtime (CPU only)**:
This stage builds the default `edge-runtime`, where `ort::Session`'s are loaded using CPU.
```bash
docker build --target "edge-runtime" .
```
> *Resulting image with ~150 Mb*

**edge-runtime-cuda (GPU/CPU)**:
This stage builds the default `edge-runtime` in a `nvidia/cuda` machine that allows loading using `GPU` or `CPU`(as fallback).
```bash
docker build --target "edge-runtime-cuda" .
```
> *Resulting image with ~2.20 GB*

Each stage needs to install the appropriated `onnx-runtime`. So in order that, the `install_onnx.sh` has updated with a 4º parameter flag `--gpu`, that will download a `cuda` version from the official [`microsoft/onnxruntime`](https://github.com/microsoft/onnxruntime/releases) repository.

#### Using GPU image:
In order to use the `gpu` image the `docker-compose` file must include the following properties for the `functions` service:

```yml
services:
  functions:
    # Built was describe before
    image: supabase/edge-runtime:latest-cuda
    # or directly by compose
    build:
      context: .
      dockerfile: Dockerfile
      target: edge-runtime-cuda

   # Required to use gpu inside the container
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: 1 # Change here if more devices are installed
              capabilities: [gpu]
```
> **IMPORTANT NOTE:** The target infrastructure must be prepared with [`NVIDIA Container Toolkit`](https://github.com/NVIDIA/nvidia-container-toolkit) to allow gpu support inside docker

---

## Final considerations:

Like I'd describe before, this is an adapted work from #368 where we spitted out only the core features that improves **ort** support for `edge-runtime`.
 
Finally, thanks for @nyannyacha that help me a loot 🙏


Dockerfile
```diff
@@ -4,42 +4,67 @@ FROM rust:1.79.0-bookworm as builder
...
-RUN objcopy --strip-debug \
-    --add-gnu-debuglink=/root/edge-runtime.debug \
-    /root/edge-runtime
-
-RUN ./scripts/install_onnx.sh $ONNXRUNTIME_VERSION $TARGETPLATFORM /root/libonnxruntime.so
-RUN ./scripts/download_models.sh
+  GIT_V_TAG=${GIT_V_VERSION} cargo build --profile ${PROFILE} --features "${FEATURES}" && \
+  cargo strip && \
+  mv /usr/src/edge-runtime/target/${PROFILE}/edge-runtime /root
+
```
nyannyacha, line 23: It looks like some contents in the main branch have been reverted.

nyannyacha:
> 👷 It's also a base foundation for an owned onnx runtime that can be integrated direclty with https://github.com/huggingface/transformers.js/pull/947 library, that will allow a better inference without the needed of coupling models like gte-small on edge-runtime image.

If so, are you also preparing other PR after this PR has been merged?
Overall, the PR looks good. 😁

nyannyacha:
Anyway, I'll be testing this locally with k6 soon.
If there are any issues, I'll let you know. 😋

Dockerfile
kallebysantos: You're right, I did miss theses `objcopy` things 🤦

nyannyacha (approved):
LTGM 👍

I've loadtested locally with k6 and it seems to be working fine.

Many thanks for your time and effort in bringing this feature to us!
I look forward to seeing you again in your follow-up PR.

cc @laktek

PR #443, opened 2024-11-12
fix(sb_ai): Sessions are dropped while still in use

kallebysantos:

Bug fix

The `cleanup` method can't verify if sessions created via `op_sb_ai_ort_init_session()` are still in use.

## What is the new behavior?

This PR fixes this issue, by ensure that sessions created from `op_sb_ai_ort_init_session()` will have its reference stored in Worker's `OpState`. This guarantees that `cleanup` method will don't drop `Sessions` attached to an active worker.

Also was introduced a better error handling when it can't found a session.

## Additional context

**Solved:**
image
> *Trying to retrieve a session that was dropped while the worker was still consuming it*

### DIFF
--- a/crates/sb_ai/onnxruntime/mod.rs
+++ b/crates/sb_ai/onnxruntime/mod.rs
@@ -3,18 +3,28 @@ pub(crate) mod onnx;
 pub(crate) mod session;
 mod tensor;
 
-use std::{borrow::Cow, collections::HashMap};
+use std::{borrow::Cow, cell::RefCell, collections::HashMap, rc::Rc, sync::Arc};
 
 use anyhow::{anyhow, Result};
-use deno_core::op2;
+use deno_core::{op2, OpState};
 
 use model_session::{ModelInfo, ModelSession};
+use ort::Session;
 use tensor::{JsTensor, ToJsTensor};
 
 #[op2]
 #[to_v8]
-pub fn op_sb_ai_ort_init_session(#[buffer] model_bytes: &[u8]) -> Result<ModelInfo> {
+pub fn op_sb_ai_ort_init_session(
+    state: Rc<RefCell<OpState>>,
+    #[buffer] model_bytes: &[u8],
+) -> Result<ModelInfo> {
+    let mut state = state.borrow_mut();
     let model_info = ModelSession::from_bytes(model_bytes)?;
+
+    let mut sessions = { state.try_take::<Vec<Arc<Session>>>().unwrap_or(Vec::new()) };
+
+    sessions.push(model_info.inner());
+    state.put(sessions);
 
     Ok(model_info.info())
 }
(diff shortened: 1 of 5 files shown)


examples/main/index.ts
```diff
@@ -11,9 +11,12 @@ console.log('main function started');
 // cleanup unused sessions every 30s
 // setInterval(async () => {
 //     const { activeUserWorkersCount } = await EdgeRuntime.getRuntimeMetrics();
+//
+//     /* Since cleanup logic will look into active consumer workers we can avoid this check
```
laktek, line 15: shall we remove this code? It may lead to incorrect implementations, that would never cleanup.

crates/sb_ai/onnxruntime/mod.rs
```diff
@@ -3,19 +3,29 @@ pub(crate) mod onnx;
...
+    #[buffer] model_bytes: &[u8],
+) -> Result<ModelInfo> {
+    let mut state = state.borrow_mut();
     let model_info = ModelSession::from_bytes(model_bytes)?;
 
+    let mut sessions = { state.try_take::<Vec<Arc<Session>>>().unwrap_or_default() };
+
+    sessions.push(model_info.inner());
+    state.put(sessions);
+
```
nyannyacha, line 28: `try_take` is to remove a value from the container.
We want to keep the Vec<T> in state, so we shouldn't remove it.

Please check if the Vec<T> exists with the `has` method, and if not, put an empty Vec<T> with `put` and then add a session with `borrow_mut`.

```rust
    let mut state = state.borrow_mut();
    
    if !state.has::<Vec<meow>>() {
        state.put(Vec::<meow>::new())
    }
    
    let vec_mut = state.borrow_mut::<Vec<meow>>();
    vec_mut.push(model_info.inner());
```

nyannyacha (dismissed):

crates/sb_ai/onnxruntime/mod.rs
nyannyacha: Ah, the code makes sense.
You don't need to fix it.

examples/main/index.ts
```diff
@@ -26,173 +29,173 @@ console.log('main function started');
...
-    // 		{
-    // 			status: STATUS_CODE.OK,
-    // 			headers,
-    // 		},
-    // 	);
+  const headers = new Headers({
+    'Content-Type': 'application/json',
+  });
+
+  const url = new URL(req.url);
```
nyannyacha, line 36: indent please

nyannyacha (approved):

laktek:
Do we still need the manual cleanup of sessions from main worker? With this refactor, sessions are stored in OpState, wouldn't that drop them after isolate terminates?

examples/main/index.ts
kallebysantos: You're right, I did remove it

PR #446, opened 2024-11-19
feat: exposing `deno_cache` + `ai cache interceptor`

kallebysantos:

feature, enhancement

When using [huggingface/transformers.js](https://github.com/huggingface/transformers.js/pull/947) the model assets can't be cached. It means that each `worker` life cycle will require a new fetch of these assets.

## What is the new behavior?

image

The `deno_cache` is now exposed to Js land, so that `transformers.js` lib can use the global `caches` object to store model assets. Also an interceptor has been attached to `deno_cache`, with this we can interpect only `.onnx` requests and use it during session load.

### How it works?

By attaching an interceptor, we can filter by `.onnx` requests and return back the `url bytes` instead of fetching the model. This way `transformers.js` lib will think that its own `model's bytes` and use it during session init. 
The exposed `onnx` runtime will knows when the incoming bytes is either `url string` or `model bytes` this way it can choose to load from internal cache or memory. 

When loading from internal cache its achieve almost the same behaviour of PR #368 (Pipeline RFC)

## Final considerations:

This is an adapted work from #368 where we spitted out only the core features that improves **ort** support for `edge-runtime`.
 
Finally, thanks for @nyannyacha that help me with this proposal and code as well 🙏


crates/sb_ai/utils.rs
```diff
@@ -0,0 +1,151 @@
...
+use xxhash_rust::xxh3::Xxh3;
+
+#[instrument(fields(%kind, url = %url))]
+pub async fn fetch_and_cache_from_url(
+    kind: &'static str,
+    url: Url,
+    cache_id: Option<String>,
+) -> Result<PathBuf, AnyError> {
+    let cache_id = cache_id.unwrap_or(fxhash::hash(url.as_str()).to_string());
+    let download_dir = ort::sys::internal::dirs::cache_dir()
```
laktek, line 19: is it possible to configure this when starting the edge-runtime?

nyannyacha: resolved

nyannyacha: `SB_AI_CACHE_DIR`
https://github.com/supabase/edge-runtime/blob/9e0a5633b83066b1624e496d5ada99193802b5be/crates/sb_ai/utils.rs#L19

PR #450, opened 2024-11-25
perf(base): sent low memory notification when worker switches to retired state

nyannyacha:

Enhancement

## Description

When the worker is retired early by the supervisor, provide timing to run GC forcibly.


laktek:
prefix as a fix so it'd make a version bump

laktek (approved):

PR #453, opened 2024-11-28
fix: polishing the worker bootstrap phase and ridding annoying cache adaptor errors emitted from `transformers.js`

nyannyacha:

Miscellaneous

## Description

November cleanup.

FYI @kallebysantos , you can access the debug variable through the `internals`.

```typescript
import { internals } from 'ext:core/mod.js';

// in debug build
console.log(internals.bootstrapArgs.opts.debug); // true
```

#### What's change

* Now, `DenoRuntime` invokes the `globalThis.bootstrapSBEdge` function directly instead of invoking it through `execute_script`.
* All the arguments passed to `globalThis.bootstrapSBEdge` are objectified. They are no longer arrays, which will allow for more flexibility when we are adding flags in the future.
* Fixes an issue where annoying error messages were emitted due to an exception propagated by `cache_adaptor.js` when `transformers.js` accessed the Cache API.

### DIFF
--- a/crates/base/src/deno_runtime.rs
+++ b/crates/base/src/deno_runtime.rs
@@ -6,7 +6,7 @@ use crate::utils::json;
 use crate::utils::path::find_up;
 use crate::utils::units::{bytes_to_display, mib_to_bytes, percentage_value};
 
-use anyhow::{anyhow, bail, Error};
+use anyhow::{anyhow, bail, Context, Error};
 use arc_swap::ArcSwapOption;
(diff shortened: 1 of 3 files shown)


laktek:
Can you update the PR title and description with high-level overview of the main refactors? Would be handy in future, if we want to trace back to a change

nyannyacha:
Since the PR isn't finished yet, I'll change the title when I finish it.

laktek (approved):

PR #469, opened 2025-01-03
fix(ext/event_worker): don't use `tracing` as the log backend if `cli/tracing` feature flag is not enabled

nyannyacha:

Bug fix

## Description

`CliLogger` does not trace the log entries emitted by tracing, so if the user does not explicitly set the event worker, the edge runtime does not render these logs to stdout or stderr.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1371,6 +1371,7 @@ dependencies = [
  "graph",
  "log",
  "once_cell",
+ "sb_event_worker",
  "tokio",
  "tracing-subscriber",
 ]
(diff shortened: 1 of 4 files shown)


sweatybridge (approved):

jdmonzn:
This fix hasn't yet been made available to be used with the current version of supabase-cli - when would you be adding it there?

nyannyacha:
Hey @jdmonzn 

This fix is already included in the pre-release.
So you can get the changes by running `npm install supabase@beta`.

PR #471, opened 2025-01-09
fix: allow setting static patterns when creating a user worker

laktek:

Currently, static file patterns (ie. static files that can be added to the eszip bundles) can be specified only via CLI flag `--static`. This PR modifies create options for user worker to support providing static patterns.

### DIFF
--- a/ext/workers/lib.rs
+++ b/ext/workers/lib.rs
@@ -91,6 +91,7 @@ pub struct UserWorkerCreateOptions {
     tmp_fs_config: Option<TmpFsConfig>,
 
     context: Option<JsonMap>,
+    #[serde(default)]
     static_patterns: Vec<String>,
 }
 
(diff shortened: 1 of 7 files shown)


nyannyacha:
Everything looks fine.

However, the indention of `examples/main/index.ts` seems to be a tab.
Can you fix it? (Starting with the Deno 2.1 LTS codebase, formatting rules will be applied to all files)
https://github.com/supabase/edge-runtime/blob/69aa03b851b4729f85060cd6f6b5d57c492a599c/examples/main/index.ts#L9

ext/workers/lib.rs
```diff
@@ -91,6 +91,7 @@ pub struct UserWorkerCreateOptions {
     tmp_fs_config: Option<TmpFsConfig>,
 
     context: Option<JsonMap>,
+    static_patterns: Vec<String>,
```
nyannyacha, line 94: I think this should be `Option<Vec<T>>`

nyannyacha: Or,
```diff
    context: Option<JsonMap>,
++ #[serde(default)]
    static_patterns: Vec<String>,
```

nyannyacha (approved):
LGTM

PR #472, opened 2025-01-13
fix: expose Supabase namespace to main worker

laktek:

Expose Supabase API to be used in main worker.


nyannyacha (approved):
LGTM

PR #473, opened 2025-01-14
Fix static file paths and race conditions

laktek:

1. If an entrypoint path is defined, use that as the root for static files
2. Don't spawn IO tasks in the current runtime


nyannyacha (approved):
LGTM

PR #483, opened 2025-02-02
fix(ext/ai): Invalid tensor length causing Segmentation Fault

kallebysantos:

Bug fix, CRITICAL

If a `Tensor` received from JS land haves `data.length` not compatible with the `dims`  then when feeding the `Session` a **Segmentation Fault: 11** error will be panic!.

### How to reproduce it:
The tensors is successfully handled by `rust` as well `JsTensor::extract_ort_input` but error will be panic when feed this values into the `Session`.
```js
const t1 = new Tensor("float32", [], [1] ); // This one always will result in panic!
const t2 = new Tensor("float32", [0.1, 0.2, 0.3], [1, 2] ); // Session may safely throws while validating inputs
// ...
await core.ops.op_sb_ai_ort_run_session("123", { t1, t2 });
```
```bash
./scripts/run.sh: line 13: 12963 Segmentation fault: 11  RUST_BACKTRACE=full ./target/debug/edge-runtime "$@" start --main-service ./examples/main --event-worker ./examples/event-manager
```

## What is the new behavior?

Following the [ort api reference](https://docs.rs/ort/2.0.0-rc.2/ort/type.TensorRefMut.html#method.from_raw):  *The 'N' of elements 'T' must be equal to the Tensor's 'dims product'*.
So we must prevent it by checking if the `Tensor` length reflects in its `dim shape`

### DIFF
--- a/ext/ai/onnxruntime/tensor.rs
+++ b/ext/ai/onnxruntime/tensor.rs
@@ -108,6 +108,15 @@ impl JsTensor {
     pub fn extract_ort_tensor_ref<'a, T: IntoTensorElementType + Debug>(
         mut self,
     ) -> anyhow::Result<ValueRefMut<'a, DynValueTypeMarker>> {
+        let expected_length = self.dims.iter().product::<i64>() as usize;
+        let current_length = self.data.len() / size_of::<T>();
+
+        if current_length != expected_length {
+            return Err(anyhow!(
+                "invalid tensor length! got '{current_length}' expect '{expected_length}'"
+            ));
+        };
+
         // Same impl. as the Tensor::from_array()
         // https://github.com/pykeio/ort/blob/abd527b6a1df8f566c729a9c4398bdfd185d652f/src/value/impl_tensor/create.rs#L170
         let memory_info = MemoryInfo::new(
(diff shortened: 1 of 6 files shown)


laktek (approved):

laktek:
@nyannyacha can you take a look?

ext/ai/onnxruntime/tensor.rs
```diff
@@ -200,3 +210,64 @@ impl ToJsTensor {
...
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_ort_tensor_extract_ref() {
+        // region: v8-init
+        // ref: https://github.com/denoland/deno_core/blob/490079f6b5c9233f476b0a529eace1f5b2c4ed07/serde_v8/tests/magic.rs#L23
+        let platform = v8::new_unprotected_default_platform(0, false).make_shared();
+        v8::V8::initialize_platform(platform);
+        v8::V8::initialize();
```
nyannyacha, line 224: The v8 platform must be initialized only once while the process is running.
There is only one unit test, so there would be no problem... but this will not be the case if there are two or more.

So please use std::Once to configure it to be initialized only once.

https://github.com/denoland/deno_core/blob/490079f6b5c9233f476b0a529eace1f5b2c4ed07/serde_v8/utilities/lib.rs#L14-L33

nyannyacha (changes requested):

ext/ai/onnxruntime/tensor.rs
nyannyacha: @kallebysantos

nyannyacha (approved):

PR #486, opened 2025-02-07
Code coverage

laktek:

This adds github actions to use Traupalin to generate code coverage report and send it to Coveralls.


nyannyacha (approved):

nyannyacha:
The failure of the last test appears to be due to a gateway timeout of the model data endpoint.

> InvalidWorkerResponse: event loop error: Error: Gateway timeout error occurred while trying to load file: "https://huggingface.co/Xenova/vit-base-patch16-224/resolve/main/onnx/model.onnx".

PR #503, opened 2025-02-26
fix: add integration tests for commonjs and enhance entrypoint discovery

nyannyacha:

Chore, Enhancement

## Description

This PR includes the following changes
* Added integration tests for commonjs
* Added examples for commonjs
* Added an integration test to check if `importMap` field in `deno.json` is recognized correctly
* Makes the main field of package.json to be recognized

Closes FUNC-180
Closes FUNC-162


laktek:
Do we also support files with `.cjs` extension? https://docs.deno.com/runtime/fundamentals/node/#use-.cjs-extension If so, good to add a test case for it too.

laktek (approved):

nyannyacha:
@laktek I tried to add support for it, but eszip doesn't support it.
https://github.com/denoland/eszip/blob/28e7a85a089e1d87ca457ca4a590718f68894159/src/v2.rs#L1406-L1445

PR #514, opened 2025-03-13
fix(deno_facade): implement `get_host_defined_options` member on `EmbeddedModuleLoader`

### LINKED ISSUE #513: deno 2.1 develop: node compat issue
Hi @nyannyacha, while trying out the development build I encountered the following

## Describe the bug

Trying out `v1.68.0-develop.7` we encountered `ReferenceError: Buffer is not defined` from the `postgres` package which indicates that node compatibility is not given.
Manually setting `unstable: ["node-globals"]` in deno.json did not help in contrary to using deno directly.

## To Reproduce

Use `v1.68.0-develop.7`
Rerference e.g. `postgres` or anything using node globals like Buffer
`supabase functions serve`

## Expected behavior

Either working directly or with the workaround in deno.json proposed in https://github.com/denoland/deno/issues/27693

nyannyacha:

Bug fix

## Description

Deno now uses a v8 API-based callback approach instead of a heuristic based on stack traces to distinguish the current execution context.
This callback is defined in the ModuleLoader trait and should have been implemented in EmbeddedModuleLoader, but was not. As a result, the current execution context was locked to Deno, which caused it to be unable to get some special keys (global, Buffer, etc...) that only exist on Node.

Fixes #513

### DIFF
--- a/crates/base/test_cases/issue-513/deno.json
+++ b/crates/base/test_cases/issue-513/deno.json
@@ -1,0 +1,6 @@
+{
+  "workspace": [],
+  "imports": {
+    "postgres": "npm:postgres@^3.4.5"
+  }
+}
--- a/crates/base/test_cases/issue-513/index.ts
+++ b/crates/base/test_cases/issue-513/index.ts
@@ -1,0 +1,10 @@
+import postgres from "postgres";
+
+const sql = postgres({/* options */});
+console.log(sql);
+
+export default {
+  fetch() {
+    return new Response(null);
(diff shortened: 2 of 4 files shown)


nyannyacha:
https://github.com/supabase/edge-runtime/actions/runs/13836155081/job/38711564880?pr=514#step:11:795 is a temporary error due to 503 status code from https://httpbin.org/stream/20.

laktek (approved):

PR #519, opened 2025-03-20
feat: support importing WebAssembly modules

nyannyacha:

Feature

## Description

This PR adds support for importing WebAssembly modules.
Now, there are no longer import errors when given snippets like the following:

math.wasm
```
<raw>
```
index.ts
```typescript
import { add } from "./math.wasm";

const three = add(1, 2) === 3;

export default {
  fetch() {
    return new Response(three ? "meow" : "woem");
  },
};
```

---

output
```
meow
```

Closes FUNC-196


laktek (approved):
🔥 this is awesome

PR #539, opened 2025-05-10
feat(ai): llm provider

kallebysantos:

feature, refactor

## What is the current behaviour?

Current the `Session` only supports self-hosted Ollama or some OpenAI like provider - with no way to specify the API key.

## What is the new behaviour?

This PR applies some refactors in the `ai` module to support an unified LLM provider API, this way it can be easily extended to new providers as well exporting a more standardised output format.

TODO! - it also unifies log events irrespective of the provider by following the OpenAI specs.
TODO! - Support for `TogetherAI`, `Amazon Bedrock`, `Google Gemini?`, `Anthropic?`


kallebysantos:
bot reply

kallebysantos:
bot reply

Issue #546, opened 2025-05-26
Binding to IPv6 ports prevents start


BenIsenstein:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

The edge runtime server fails to run when receiving an IPv6 host from the `--ip` option. Ex. `"::"`, `"::1"`
The error logged is `Error: invalid IPv4 address syntax`

## To Reproduce

```bash
git clone https://github.com/BenIsenstein/supabase-edge-runtime-ipv6-bug && \
cd supabase-edge-runtime-ipv6-bug && \
docker build -t supabase-edge-runtime-ipv6-bug . && \
docker run -p 9000:9000 supabase-edge-runtime-ipv6-bug
```

## Expected behavior

The server is able to run listening for IPv6 connections.

## Screenshots

No screenshots.

## System information

- OS: MacOS working in Docker
- Edge Runtime Version: v1.67.4

## Additional context

None

kallebysantos:
Hi guys 💚

I confirm that I was able to reproduce the error, tested both from the instructions provided by @BenIsenstein as well by running directly from source code with `./scripts/run.sh`

### Logs

<details>
<summary>Running from the code snippet provided by the author</summary>

**Error:** `Error: invalid IPv4 address syntax`, probably handled from Js land

```
❯ git clone https://github.com/BenIsenstein/supabase-edge-runtime-ipv6-bug && \
      cd supabase-edge-runtime-ipv6-bug && \
      docker build -t supabase-edge-runtime-ipv6-bug . && \
      docker run -p 9000:9000 supabase-edge-runtime-ipv6-bug
Cloning into 'supabase-edge-runtime-ipv6-bug'...
remote: Enumerating objects: 9, done.
remote: Counting objects: 100% (9/9), done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 9 (delta 0), reused 9 (delta 0), pack-reused 0 (from 0)
Receiving objects: 100% (9/9), done.
[+] Building 0.8s (7/7) FINISHED                                                                                                                                                                     docker:orbstack
 => [internal] load build definition from Dockerfile                                                                                                                                                            0.1s
 => => transferring dockerfile: 225B                                                                                                                                                                            0.0s
[...]

PR #547, opened 2025-05-27
fix(cli): ip address parsing to support both ipv4 and ipv6

kallebysantos:

Bug fix

IPv6 addresses causes error

## What is the new behavior?

solves #546


nyannyacha (approved):
LGTM!

PR #551, opened 2025-06-14
deps: Add cargo-shear

hkrutzer:

Remove unnecessary dependencies using Cargo Shear

N/a

## What is the new behavior?

N/a

## Additional context

https://crates.io/crates/cargo-shear

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,20 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - run: ./scripts/clippy.sh
 
+  cargo-shear:
+    if: github.event.pull_request.draft == false
+    name: 'cargo shear'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: rustup show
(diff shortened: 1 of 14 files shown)


laktek:
@nyannyacha we do need cargo-shear yeah?

nyannyacha:
@laktek Well... I'm not sure if we need this.

It's nice to have Catgo.toml leave only the dependencies it needs, but as far as I know, the tool doesn't detect dependencies used by code generated by proc macros, etc. (We'd eventually have to use the nightly version of the Rust toolchain to detect this as described in its docs).

Even if the unused dependencies are in Cargo.toml, this isn't necessary, as rustc's opt passes will elliminates all unused code.

hkrutzer:
Well, while dependencies used by generated code are not detected that is a hypothetical, while compiling unused dependencies, taking compile time and disk space, are a current tangible cost.

nyannyacha:
I think your point is mostly correct.

Unfortunately, I am not prioritizing this at the moment, so I think it would be best to leave it closed for now.
will review this PR in depth when the time comes.

Thank you for your efforts!

PR #557, opened 2025-07-14
feat: add v2.2.0 `node:sqlite` support

nyannyacha:

Feature

## Description

Brings in the SQLite extension that exists in Deno v2.2.0 to edge-runtime.

### DIFF
--- a/ext/node/ops/sqlite/database.rs
+++ b/ext/node/ops/sqlite/database.rs
@@ -8,6 +8,8 @@ use deno_core::anyhow;
 use deno_core::anyhow::anyhow;
 use deno_core::op2;
 use deno_core::GarbageCollected;
+use deno_core::OpState;
+use deno_permissions::PermissionsContainer;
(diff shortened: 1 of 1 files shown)


ext/node/ops/sqlite/database.rs
```diff
@@ -0,0 +1,166 @@
...
+  // This method should only be used when the database is not opened
+  // via the constructor. An exception is thrown if the database is
+  // already opened.
+  #[fast]
+  fn open(&self) -> Result<(), anyhow::Error> {
+    if self.conn.borrow().is_some() {
+      return Err(anyhow!("Database is already open"));
+    }
+
+    let db = rusqlite::Connection::open(&self.location)?;
```
laktek, line 90: looks like this version doesn't check if deno process has permission to open file at the given location. Can we add the perm check?

laktek (approved):

PR #582, opened 2025-08-11
feat: add early shutdown when no requests received for X seconds

nyannyacha:

Feature

## Description
...WIP


laktek (approved):

farzd:
Does this have any effect if requestAbsentTimeoutMs is not set? @nyannyacha 
My edge functions have started shutting down mid tasks 

Screenshot 2025-08-19 at 14 20 02

laktek:
@farzd Are you using Background tasks to run the function? If so, it wouldn't shutdown early. https://supabase.com/docs/guides/functions/background-tasks

Also from the log events, I cannot see a shutdown event. When a function is shutdown there would be a shutdown log event which shows the reason for shutdown. The early shutdowns will show as "EarlyDrop" as the reason.

farzd:
Thanks for the reply @laktek . I'm not using background Tasks. 
I have another edge function that returns early and fires off this one. I assumed on Pro plan you get 400 seconds so it would stay alive until that ?

(also please note, i only started having these issues as of recent, wasn't getting early drops before)

image

```
 [
  {
    "boot_time": null,
    "cpu_time_used": 153,
    "deployment_id": "xxxx",
    "event_type": "Shutdown",
    "execution_id": "aee2c2d4-8beb-484b-90d2-2e572dc0fdb5",
    "function_id": "1a0eb3e7-b021-4f40-b80e-f43ed4e5ba14",
    "level": "log",
    "memory_used": [
      {
        "external": 10811056,
        "heap": 12861832,
        "mem_check_captured": null,
        "total": 23672888
      }
    ],
    "project_ref": "xxxx",
    **"reason": "EarlyDrop",**
    "region": "eu-west-2",
    "served_by": "supabase-edge-runtime-1.68.0-develop.36 (compatible with Deno v2.1.4)",
    "timestamp": "2025-08-19T13:52:20.101Z",
    "version": "18"
  }
]
```

laktek:
> I'm not using background Tasks.

>  I assumed on Pro plan you get 400 seconds so it would stay alive until that ?

Pro plan gives you 400s, but if the worker (isolate) becomes idle it'll shutdown. The way we detect if the isolate is idle is by checking if it has pending requests or active background tasks. So if you update your code to run the long-running task as a background task it'll ensure to run for 400s.

> (also please note, i only started having these issues as of recent, wasn't getting early drops before)

We always had this behavior, but recently, we refactored the code so it's applied based on the load in each region. This ensures all users get a fair amount of resources, which is probably why you started having these issues recently.

For further support, I'd recommend opening a support ticket.

farzd:
Thanks @laktek

PR #608, opened 2025-09-12
feat: Add configurable functions path via SUPABASE_FUNCTIONS_PATH environment variable

kshivang:
## Problem

Currently, the edge-runtime has a hardcoded path for function compilation at `/var/tmp/sb-compile-<executable-name>`. This creates compatibility issues with deployment scenarios that require custom paths, particularly:

1. **Kubernetes deployments** using community Helm charts that mount functions at different paths (e.g., `/home/deno/functions`)  
2. **Restricted environments** where `/var/tmp` may not be writable or available
3. **Custom Docker deployments** that prefer different directory structures

## Solution

This PR adds a configurable `SUPABASE_FUNCTIONS_PATH` environment variable that allows users to override the default function compilation path while maintaining full backward compatibility.

### Changes Made:

1. **Modified `standalone.rs`**: Added environment variable check for `SUPABASE_FUNCTIONS_PATH` with fallback to existing behavior
2. **Updated `README.md`**: Added documentation for the new environment variable
3. **Maintained backward compatibility**: Existing deployments continue to work unchanged

### Code Changes:

```rust
let root_path = match std::env::var("SUPABASE_FUNCTIONS_PATH") {
  Ok(custom_path) => PathBuf::from(custom_path),
  Err(_) => {
    // Default behavior: use platform-specific temp directory
    if cfg!(target_family = "unix") {
      PathBuf::from("/var/tmp")
    } else {
      std::env::temp_dir()
    }
    .join(format!("sb-compile-{}", current_exe_name))
  }
};
```

### Default behavior (unchanged):
- Unix/Linux: `/var/tmp/sb-compile-edge-runtime`
- Windows: `%TEMP%\sb-compile-edge-runtime`

### Custom path:
```bash
export SUPABASE_FUNCTIONS_PATH=/home/deno/functions
./edge-runtime start --main-service /path/to/function
```

### Kubernetes example:
```yaml
env:
  - name: SUPABASE_FUNCTIONS_PATH
    value: "/home/deno/functions"
```

## Testing

- ✅ Maintains backward compatibility (no environment variable set)
- ✅ Respects custom path when `SUPABASE_FUNCTIONS_PATH` is set
- ✅ Documentation updated

## Benefits

1. **Kubernetes compatibility**: Enables edge functions to work with community Helm charts
2. **Deployment flexibility**: Allows custom paths for different deployment scenarios  
3. **Zero breaking changes**: Existing deployments continue working without modification
4. **Security**: Enables deployment in environments with restricted `/var/tmp` access

## Related Issues

- Addresses hardcoded path incompatibility reported in supabase-community/supabase-kubernetes#117
- Enables edge functions support for community Kubernetes deployments

---

**Reported-by**: kshivang  
**Tested with**: edge-runtime v1.69.6+

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,14 @@ The edge runtime can be divided into two runtimes with different purposes.
   - Limits are required to be set such as: Memory and Timeouts.
   - Has access to environment variables explictly allowed by the main runtime.
 
+## Configuration
+
+### Environment Variables
+
+- `SUPABASE_FUNCTIONS_PATH` - Custom path for function compilation directory. 
+  - Default: `/var/tmp/sb-compile-<executable-name>` on Unix/Linux, `%TEMP%\sb-compile-<executable-name>` on Windows
+  - Example: `SUPABASE_FUNCTIONS_PATH=/home/deno/functions` (useful for Kubernetes deployments)
+
 ## Developers
 
 To learn how to build / test Edge Runtime, visit [DEVELOPERS.md](DEVELOPERS.md)
--- a/crates/deno_facade/module_loader/standalone.rs
+++ b/crates/deno_facade/module_loader/standalone.rs
@@ -579,12 +579,18 @@ pub async fn create_module_loader_for_eszip(
   .transpose()?
   .unwrap_or_default();
 
-  let root_path = if cfg!(target_family = "unix") {
-    PathBuf::from("/var/tmp")
-  } else {
-    std::env::temp_dir()
-  }
-  .join(format!("sb-compile-{}", current_exe_name));
+  let root_path = match std::env::var("SUPABASE_FUNCTIONS_PATH") {
+    Ok(custom_path) => PathBuf::from(custom_path),
+    Err(_) => {
+      // Default behavior: use platform-specific temp directory
+      if cfg!(target_family = "unix") {
+        PathBuf::from("/var/tmp")
+      } else {
+        std::env::temp_dir()
+      }
+      .join(format!("sb-compile-{}", current_exe_name))
+    }
+  };
 
   let node_modules = metadata.node_modules()?;
   let root_dir_url =


kallebysantos:
Hi @kshivang 💚
Is `install_onnx.sh` related with the main thread of this PR? or you just took the opportunity to refactor it as well?
May you please add a note of this change on your PR description? - for documenting

kshivang:
sure I will

kshivang:
Hey @kallebysantos I removed other commit from this PR, keeping just first commit that is relevant for this fix. For onnx I will create separate PR, as it is not relevant for this PR

nyannyacha:
Hi @kshivang 😁

I think this PR needs more context to understand the background.
Also looked at the issue you opened [here](https://github.com/supabase-community/supabase-kubernetes/issues/117), but I don't quite understand why this change is needed.

I'll quote the error from the issue you opened in the k8s repo.
```
worker boot error: [ERR_MODULE_NOT_FOUND] Cannot find module 
'file:///var/tmp/sb-compile-edge-runtime/node_modules/localhost/@supabase/realtime-js/...'
```

This error message may be confusing... because the path it indicates appears to point to the host's /var/tmp directory.
But, this path is actually mapped to edge-runtime's in-memory VFS, not the host's /var/tmp directory. Therefore, it has no interaction whatsoever with the corresponding path on the host.

I'm not quite sure, but did the function you used for this issue work properly when loaded in the edge-runtime after applying this PR?

And if it's okay with you, could I get the function (or project) used to reproduce this issue?

kastakhov:
Hi @nyannyacha, seems I have found why this error is occurred. In k8s when we mount configMap as a volume and don't specify a file which need to be mounted, we got all files from the config map as a symlinks, something like next:
```
root@debian-dummy-2:/# ls -la /home/deno/functions/main/
total 0
drwxrwxrwx. 3 root root 76 Oct  2 15:13 .
drwxr-xr-x. 3 root root 18 Oct  2 15:13 ..
drwxr-xr-x. 2 root root 22 Oct  2 15:13 ..2025_10_02_15_13_34.2409747743
lrwxrwxrwx. 1 root root 32 Oct  2 15:13 ..data -> ..2025_10_02_15_13_34.2409747743
lrwxrwxrwx. 1 root root 15 Oct  2 15:13 index.ts -> ..data/index.ts
```
And per my understanding, the only symlink itself is mapped to the edge-runtime's in-memory VFS, but not the actual file recursively.

@kshivang I also shared the fix in your original issue(https://github.com/supabase-community/supabase-kubernetes/issues/117)

kshivang:
> Hi @kshivang 😁
> 
> I think this PR needs more context to understand the background. Also looked at the issue you opened [here](https://github.com/supabase-community/supabase-kubernetes/issues/117), but I don't quite understand why this change is needed.
> 
> I'll quote the error from the issue you opened in the k8s repo.
> 
> ```
> worker boot error: [ERR_MODULE_NOT_FOUND] Cannot find module 
> 'file:///var/tmp/sb-compile-edge-runtime/node_modules/localhost/@supabase/realtime-js/...'
> ```
> 
> This error message may be confusing... because the path it indicates appears to point to the host's /var/tmp directory. But, this path is actually mapped to edge-runtime's in-memory VFS, not the host's /var/tmp directory. Therefore, it has no interaction whatsoever with the corresponding path on the host.
> 
> I'm not quite sure, but did the function you used for this issue work properly when loaded in the edge-runtime after applying this PR?
> 
> And if it's okay with you, could I get the function (or project) used to reproduce this issue?

The main issue was symbolic, as @kastakhov mentioned. Although his PR solves the problem for now, for a secure Kubernetes environment, such as Google Kubernetes Autopilot, the path should be configurable. I was not getting an error after this PR because I was able to give the edge cloud function the right path.

PR #614, opened 2025-09-20
Fix scripts/run.sh for macOS

gofenix:
This pull request is to fix the issue of executing run.sh on macOS.

### DIFF
--- a/scripts/run.sh
+++ b/scripts/run.sh
@@ -1,13 +1,18 @@
 #!/usr/bin/env bash
 set -e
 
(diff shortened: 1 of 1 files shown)


kallebysantos:
Hi @gofenix 💚
May you add the issue log into PR description? I'm using macOS but didn't found any issue.
Its help with documenting and future user searching

gofenix:
I encountered an error earlier, but then found that changing this script fixed it. However, when I tried again just now, I couldn't reproduce the problem. Sorry about that.

PR #628, opened 2025-10-24
auto cleanup ts from test on drop

johnstonmatt:

test related fix

- running `./scripts/test.sh` can result in a failing test which leaves a stray typescript file `./test_cases/eszip-source-test.ts` which then results in a git diff and a failing format check

## What is the new behavior?

- test now automatically cleans up the typescript file as intended


nyannyacha:
Please ignore the 4 tests that failed in CI.
I know why they failed and will fix this issue in the #624.

Issue #542, opened 2025-05-19
Third Party Auth JWT not compatible with Edge Function


PBJT2:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Following the guide at [Supabase Auth0 Integration Docs](https://supabase.com/docs/guides/auth/third-party/auth0).  
I was not able to use the JWT issued by Auth0 to invoke my edge function.
I was able to use the JWT issued by Auth0 to query the database.

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Follow the guide at [Supabase Auth0 Integration Docs](https://supabase.com/docs/guides/auth/third-party/auth0) for setup.
2. Attempt to invoke Edge Function with a JWT issued by Auth0

## Expected behavior

Auth0 JWT should be a valid JWT to invoke the edge function with.

## System information

- OS: Windows 11
- Version of supabase-js: 2.49.4
- Version of Node.js: 22

nyannyacha:
This seems to be an issue with Edge Functions and its JWT token verification.
It does not appear to be directly related to edge-runtime, but FYI @laktek.

shawnesquivel:
This relates to https://github.com/orgs/supabase/discussions/34988  - would love to get some supabase team commentary on this as the docs are a bit vague on  Clerk <-> Supabase inteagration regarding the jwt flag!

nascosto:
Still an issue

oskarsvard-resly:
Yes, still an issue for me to.

nascosto:
Still an issue

kallebysantos:
Hi @nascosto 💚
May you help me with an example project?

> EDIT: I got one from https://github.com/clerk/clerk-supabase-nextjs

can someone confirm that if error logs is looking like that?
```
2025-11-12T15:33:06.206214977Z  - http://127.0.0.1:54321/functions/v1/test_auth
2025-11-12T15:33:06.206217685Z Using supabase-edge-runtime-1.69.15 (compatible with Deno v2.1.4)
2025-11-12T15:33:14.349071910Z TypeError: Key for the RS256 algorithm must be of type CryptoKey. Received an instance of Uint8Array
[...]

PR #642, opened 2025-11-12
chore: jwt examples

kallebysantos:

docs update

## What is the new behaviour?
Since jwt verification will be deprecated, users should use `--no-verify-jwt` and do verification by their selves. 

This PR adds some templates for jwt token verification.

## Additional context

- related with https://github.com/orgs/supabase/discussions/34988
- partially solves https://github.com/supabase/edge-runtime/issues/542

### DIFF
--- a/examples/_shared/jwt/clerk.ts
+++ b/examples/_shared/jwt/clerk.ts
@@ -1,9 +1,9 @@
-// Clerk Provider JWT verification
+// Clerk as a third-party provider alongside Supabase Auth.
 // Use this template to validate tokens issued by Clerk integration
-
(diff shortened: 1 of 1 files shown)


nyannyacha:
Adding examples is always welcome! Good job @kallebysantos !

btw this is just adding an example, right? So to avoid version changes during merging, we should change the `feat` prefix to `chore`. Is that okay?

kallebysantos:
Sure! thank you for explain me that 💚
Theses examples should help with the issues I did linked, you can find more context [here](https://supabase.slack.com/archives/C03S3SW9485/p1762964868692609)

Issue #645, opened 2025-11-26
Bug: ArrayBuffer allocator doesn't rollback count on failed allocation


max-lt:
- [ ] I confirm this is a bug with Supabase, not with my own application. 
- [ ] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

In [ext/runtime/external_memory.rs](https://github.com/supabase/edge-runtime/blob/main/ext/runtime/external_memory.rs), when an allocation exceeds the limit, the count is incremented but never rolled back on failure.

  allocate and allocate_uninitialized
```
  unsafe extern "C" fn allocate(allocator: &CustomAllocator, n: usize) -> *mut c_void {
      allocator.count.fetch_add(n, Ordering::SeqCst);  // count += n

      let count_loaded = allocator.count.load(Ordering::SeqCst);

      if count_loaded > allocator.max {
          return std::ptr::null::<*mut [u8]>() as *mut c_void;  // returns without rollback
      }
      // ...
  }
```
  reallocate
```
  unsafe extern "C" fn reallocate(
    allocator: &CustomAllocator,
    prev: *mut c_void,
    oldlen: usize,
    newlen: usize,
  ) -> *mut c_void {
    allocator
      .count
      .fetch_add(newlen.wrapping_sub(oldlen), Ordering::SeqCst);  // count += (newlen - oldlen)

    let count_loaded = allocator.count.load(Ordering::SeqCst);

    if count_loaded > allocator.max {
      return std::ptr::null::<*mut [u8]>() as *mut c_void;  // returns without rollback
    }
    // ...
  }
```

  Impact

  Attempt 1: allocate(100MB), limit=50MB → count=100MB → FAIL → count stays 100MB
  Attempt 2: allocate(1KB)              → count=100MB+1KB → FAIL (even though 1KB < 50MB)

  All future allocations fail permanently, even small legitimate ones.

  Fix
```
  // allocate / allocate_uninitialized
  if count_loaded > allocator.max {
      allocator.count.fetch_sub(n, Ordering::SeqCst);  // rollback
      return std::ptr::null::<*mut [u8]>() as *mut c_void;
  }

  // reallocate
  if count_loaded > allocator.max {
      allocator.count.fetch_sub(newlen.wrapping_sub(oldlen), Ordering::SeqCst);  // rollback
[...]

PR #646, opened 2025-11-26
fix(ext/runtime): rollback memory count #645

max-lt:
When allocate, allocate_uninitialized, or reallocate exceeded the memory
  limit, the count was incremented but never rolled back on failure. This
  caused all future allocations to fail permanently, even small legitimate
  ones.

Bug fix for #645


nyannyacha:
Thank you! @max-lt

PR #655, opened 2026-01-19
fix(devcontainer): add zstd dependency to fix build

luizfelmach:

Bug fix.

The devcontainer build fails due to a missing `zstd` dependency required during the build process.

## What is the new behavior?

The devcontainer builds successfully after adding `zstd` to the Dockerfile dependencies.

## Additional context

This change ensures the development environment works out of the box without requiring manual dependency installation.


kallebysantos:
Thx for contributing @luizfelmach 💚
I confirm that it solved issues while building on ARM

cc @nyannyacha

PR #662, opened 2026-02-20
fix(ext/node): brotli async variant of brotliDecompress

kallebysantos:

Bug fix

Node `zlib.brotliDecompress` doesn't support async wrappers.

## What is the new behavior?

Applied the same patches from upstream: https://github.com/denoland/deno/pull/27815


nyannyacha:
Ignoring the 4 ort-related test failures that failed on the last attempt (because they are flaky due to rate limits on the hg endpoint)

PR #670, opened 2026-03-03
feat: add request-local rate limiter

nyannyacha:

Feature

Towards FUNC-479

## Description

Adds a request-local outbound HTTP rate limiter that limits the number of outbound HTTP requests a worker can make, keyed by the W3C trace context. This prevents runaway circular fetch chains (`A → B → A → …`) and recursive self-calling loops from running indefinitely.

## How it works

When an inbound request arrives, the runtime extracts the trace ID from the `traceparent` header and stores it in an `AsyncVariable`. Every subsequent outbound HTTP request made by that worker — whether through `fetch` or `node:http` — reads this value and increments a counter against it.

- **Traced requests** (inbound request carries a `traceparent` header): budget is tracked per trace ID (`local` budget). All outbound requests within the same trace share one counter.
- **Untraced requests** (no `traceparent` header): budget is tracked by the `key` supplied at worker creation time (`global` budget). All user workers of the same function share one counter regardless of how many are running (per process).

When the budget is exhausted, `fetch()` and `node:http` requests throw `Deno.errors.RateLimitError`.

## Usage

Configure `traceRateLimitOptions` when creating a user worker from the main worker script. To have `traceparent` automatically propagated into outbound requests (so downstream workers can participate in the same trace), also enable the OTel `TraceContext` propagator.

```ts
// main worker
const worker = await EdgeRuntime.userWorkers.create({
  servicePath,
  // ...

  // Required for automatic traceparent propagation to outbound requests.
  otelConfig: {
    tracing_enabled: true,
    propagators: ["TraceContext" /*, "Baggage" */],
  },

  traceRateLimitOptions: {
    // Stable identifier shared across all user workers of the same function.
    // Used as the rate-limit key for untraced requests.
    // Typically set to the servicePath (or deployment id).
    key: servicePath,
    rules: [
      {
        matches: ".*", // regex matched against the outbound request URL
        ttl: 60, // budget reset interval in seconds
        budget: {
          local: 20, // max outbound requests per trace ID within the TTL
          global: 50, // max outbound requests across all untraced calls within the TTL
        },
      },
    ],
  },
});

```

## Newly introduced CLI Flags
| Flag | Description | Default |
|------|-------------|---------|
| `--rate-limit-table-cleanup-interval <SECONDS>` | How often expired rate-limit entries are swept from the shared table | `60` |

### DIFF
--- a/vendor/deno_fetch/26_fetch.js
+++ b/vendor/deno_fetch/26_fetch.js
@@ -395,8 +395,7 @@ function fetch(input, init = { __proto__: null }) {
         return;
       }
 
-      // Rate limit check — placed after abort so cancelled requests don't
-      // consume budget.
+      // Rate limit check.
       const traceId = internals.getRequestTraceId?.();
       const isTraced = traceId !== null && traceId !== undefined;
       const rlKey = isTraced ? traceId : "";
@@ -412,6 +411,7 @@ function fetch(input, init = { __proto__: null }) {
         reject(new Deno.errors.RateLimitError(msg));
         return;
       }
+
       // 7.
       let responseObject = null;
       // 9.


jasonradford115-ui (commented):
Y

jasonradford115-ui:
> Y

PR #700, opened 2026-05-06
chore(nix): fix macos build missing shared libraries

kallebysantos:

Bug fix

macos build is missing some shared libraries

## What is the new behavior?

Changing the way nix libs are included


reaganmcf:
Any update on this? This would be really beneficial for us to test stuff locally on MacOS

PR #704, opened 2026-05-17
fix(deno_facade): allow non-MainModule resolution when referrer is "."

### LINKED ISSUE #703: Bug: require()-ing a pure-ESM module from CJS crashes the worker
# Bug report

- [X] I confirm this is a bug with Supabase, not with my own application. 
- [X] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

When a CommonJS module synchronously `require()`s a pure-ESM module (either
a local `.mjs` file or a `"type": "module"` npm package), the user worker
crashes during resolution with:

```
Error: Expected to resolve main module, got Import instead.
    at loadESMFromCJS (node:module:777)
```

## To Reproduce

Drop a minimal CJS-requires-ESM function into `examples/`:

```
examples/commonjs-require-esm/
├── package.json     { "type": "commonjs" }
├── index.js         const esm = require("./message.mjs");
│                    Deno.serve(() => new Response(esm.message));
└── message.mjs      export const message = "cjs require esm ok";
```

Start the runtime and hit the function:

```bash
./scripts/run.sh                              # main worker on :9000
curl http://localhost:9000/commonjs-require-esm
```

Expected: `cjs require esm ok`.
Actual: the user worker fails to boot with the error above.

The same happens with any real pure-ESM npm dependency
(e.g. `node-fetch@3`, `chalk@5`).

## Expected behavior

CJS → ESM `require()` should succeed, matching Node and Deno's
standalone runtime behavior. The function above should respond with
`cjs require esm ok`.

0x5457:

Bug fix

`EmbeddedModuleLoader::resolve` rejects any resolution where `referrer == "."`
unless `kind == ResolutionKind::MainModule`:

```rust
if kind != ResolutionKind::MainModule {
    return Err(generic_error(format!(
        "Expected to resolve main module, got {:?} instead.",
        kind
    )));
}
```

This breaks Deno's CJS→ESM bridge. When a CommonJS module synchronously
`require()`s a pure-ESM module, `loadESMFromCJS` re-enters `resolve` with
`ResolutionKind::Import` and the same `"."` referrer, and the worker boots
with:

```
Error: Expected to resolve main module, got Import instead.
    at loadESMFromCJS (node:module:777)
```

fix #703 

## What is the new behavior?

After the fix, `require("./message.mjs")` and `require("chalk")` (a
`"type": "module"` package) both load correctly from a CJS entrypoint.

### DIFF
--- a/crates/base/test_cases/commonjs-require-esm/index.js
+++ b/crates/base/test_cases/commonjs-require-esm/index.js
@@ -1,0 +1,7 @@
+const esm = require("./message.mjs");
+
+if (esm.message !== "cjs require esm ok") {
+  throw new Error(`unexpected ESM value: ${esm.message}`);
+}
+
+Deno.serve(() => new Response(esm.message));
--- a/crates/base/test_cases/commonjs-require-esm/message.mjs
+++ b/crates/base/test_cases/commonjs-require-esm/message.mjs
@@ -1,0 +1,1 @@
+export const message = "cjs require esm ok";
--- a/crates/base/test_cases/commonjs-require-esm/package.json
+++ b/crates/base/test_cases/commonjs-require-esm/package.json
@@ -1,0 +1,4 @@
+{
+  "workspaces": [],
+  "type": "commonjs"
+}
--- a/crates/base/tests/integration_tests.rs
+++ b/crates/base/tests/integration_tests.rs
@@ -3255,6 +3255,25 @@ async fn test_commonjs() {
 
 #[tokio::test]
 #[serial]
+async fn test_commonjs_require_esm() {
+  integration_test!(
+    "./test_cases/main",
+    NON_SECURE_PORT,
+    "commonjs-require-esm",
+    None,
+    None,
+    None,
+    (|resp| async {
+      let resp = resp.unwrap();
+      assert_eq!(resp.status().as_u16(), 200);
+      assert_eq!(resp.text().await.unwrap().as_str(), "cjs require esm ok");
+    }),
+    TerminationToken::new()
+  );
+}
+
+#[tokio::test]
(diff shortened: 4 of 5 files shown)


nyannyacha:
Hello @0x5457 😋
Thank you for submitting this.

Have you tried running `deno compile` in Deno v2.1.4?
The module loader implementation in the edge-runtime codebase reflects the EmbeddedModuleLoader implementation in the `deno compile` subcommand path of Deno v2.1.4 codebase, so it appears to behave consistently; however, `deno run` seems to work without error, which suggests an inconsistency.

I think this change is acceptable, but do you have any special comments regarding the above?

0x5457:
Hi @nyannyacha, thanks for the review!

I haven't tried deno compile on v2.1.4 directly, but I noticed that upstream Deno hit the same issue and fixed it in denoland/deno#28379. That PR removes the exact same `if kind != ResolutionKind::MainModule { ... }` guard from `EmbeddedModuleLoader::resolve` in `cli/rt/run.rs`, its description explicitly mentions "fixes a bug where require esm wasn't working in deno compile", which matches what we're seeing here.

https://github.com/denoland/deno/pull/28379/changes#diff-c30a4785f54c02866be5d07ef0ce993d0518eeea81a85344ed953de6bbcf852cL156-L167

nyannyacha:
Some tests seem a bit flaky due to third-party sources, but as the fix itself appears to be fine, so LTGM.
Thank you for your efforts! 😄

nyannyacha (approved):

PR #706, opened 2026-06-01
fix: preserve system certs when adding Mozilla roots

### LINKED ISSUE #705: DENO_TLS_CA_STORE="system,mozilla" silently drops system certs (order-dependent regression from upstream Deno)
## Summary

In [`crates/deno_facade/cert_provider.rs`](https://github.com/supabase/edge-runtime/blob/main/crates/deno_facade/cert_provider.rs), the `"mozilla"` match arm **assigns** to `root_cert_store`, overwriting it, while the `"system"` arm correctly **appends**. As a result, the documented value `DENO_TLS_CA_STORE="system,mozilla"` silently discards every cert that was just loaded from the system trust store — leaving only the embedded webpki bundle.

This is order-dependent (`"mozilla,system"` works, `"system,mozilla"` does not), and the `_` match arm even advertises `"system,mozilla"` as the allowed format in its error message — so the value most users will copy from the error string is the broken one.

## Repro

Corporate proxy (in our case Cato Networks) doing TLS MITM. Steps:

1. Set `DENO_TLS_CA_STORE=system,mozilla` in `supabase/functions/.env`
2. `docker exec` into the running `supabase_edge_runtime_*` container, copy the corporate CA into `/usr/local/share/ca-certificates/`, run `update-ca-certificates`. The bundle at `/etc/ssl/certs/ca-certificates.crt` now contains the corporate CA.
3. Restart the container so the runtime re-runs `get_root_cert_store_provider()`.
4. From inside the container, `openssl s_client -connect api.openai.com:443 -showcerts` successfully verifies the chain via the system store (proving the cert is correctly installed).
5. A Deno fetch to the same host from a function still fails with:
   ```
   TypeError: error sending request for url (https://api.openai.com/...): client error (Connect): invalid peer certificate: UnknownIssuer
   ```
6. Change the env var to `DENO_TLS_CA_STORE=system` (or `mozilla,system`) and re-create the container — fetch succeeds.

## Root cause

[cert_provider.rs](https://github.com/supabase/edge-runtime/blob/main/crates/deno_facade/cert_provider.rs):

```rust
for store in ca_stores.iter() {
  match store.as_str() {
    "mozilla" => {
      root_cert_store = deno_tls::create_default_root_cert_store();  // ← REPLACES the store
    }
    "system" => {
      let roots = load_native_certs().expect("could not load platform certs");
      for root in roots {
        root_cert_store.add((&*root.0).into())...;                   // ← APPENDS
      }
    }
    _ => bail!("... (allowed: \"system,mozilla\")"),                 // ← suggests the broken value
  }
}
```

The comment in the file links to [denoland/deno@v1.37.0 cli/args/mod.rs#L467](https://github.com/denoland/deno/blob/v1.37.0/cli/args/mod.rs#L467) as the reference implementation. Upstream Deno is **append-only** on both arms:

```rust
"mozilla" => {
  root_cert_store.add_trust_anchors(            // ← appends, not assigns
    webpki_roots::TLS_SERVER_ROOTS.iter().map(...)
  );
}
"system" => {
  let roots = load_native_certs().expect(...);
  for root in roots {
    root_cert_store.add(...)...;
  }
}
```

So the regression was introduced when porting to use `deno_tls::create_default_root_cert_store()` — that helper builds a fresh store, but the surrounding loop assumed append semantics like upstream.

## Suggested fix

Either:

**(a)** Extend rather than assign in the mozilla arm:
```rust
"mozilla" => {
  let mozilla_store = deno_tls::create_default_root_cert_store();
  root_cert_store.roots.extend(mozilla_store.roots);
}
```

**(b)** Mirror upstream by calling `add_trust_anchors` directly with `webpki_roots` (would need to depend on `webpki-roots` directly rather than going through `deno_tls`'s helper).

Either way, order-independence should be a test: `"system,mozilla"` and `"mozilla,system"` must produce the same effective trust set.

## Affected version

Reproduced on `supabase-edge-runtime-1.73.13` (also visually confirmed present on `main` at the time of this report). Supabase CLI v2.95.4 on Windows 11 + Docker Desktop.

AyaanFaisal21:

Bug fix. 

Fixes #705.

The environment variable configuration`DENO_TLS_CA_STORE=system,mozilla` loads system roots first, but then the `mozilla` branch replaces the existing `RootCertStore`, which drops the system roots that were just added.

This makes CA store behavior order-dependent: `mozilla,system` preserves both stores, while `system,mozilla` only keeps Mozilla/webpki roots.

## What is the new behavior?

The `mozilla` branch now appends Mozilla/webpki roots to the existing `RootCertStore` instead of replacing it entirely.

This preserves system roots for the documented `DENO_TLS_CA_STORE=system,mozilla` environment variable configuration and makes `system,mozilla` / `mozilla,system` produce the same effective trust-anchor set.

I also added a deterministic regression test covering both orderings without depending on the host machine’s system certificate store.

## Additional context

Verification run locally:

- `cargo fmt --check`
- `cargo test -p deno_facade ca_store_order_does_not_drop_system_roots`
- `cargo test -p deno_facade`
- `git diff --check`

### DIFF
--- a/crates/deno_facade/cert_provider.rs
+++ b/crates/deno_facade/cert_provider.rs
@@ -1,12 +1,16 @@
 use std::sync::Arc;
 
 use anyhow::bail;
-use deno::deno_tls;
 use deno::deno_tls::deno_native_certs::load_native_certs;
 use deno::deno_tls::rustls::RootCertStore;
+use deno::deno_tls::webpki_roots;
 use deno::deno_tls::RootCertStoreProvider;
 use deno_core::error::AnyError;
 use ext_runtime::cert::ValueRootCertStoreProvider;
+
+fn add_mozilla_roots(root_cert_store: &mut RootCertStore) {
+  root_cert_store.extend(webpki_roots::TLS_SERVER_ROOTS.iter().cloned());
+}
 
 pub fn get_root_cert_store_provider(
 ) -> Result<Arc<dyn RootCertStoreProvider>, AnyError> {
@@ -27,9 +31,7 @@ pub fn get_root_cert_store_provider(
 
   for store in ca_stores.iter() {
     match store.as_str() {
-      "mozilla" => {
-        root_cert_store = deno_tls::create_default_root_cert_store();
-      }
+      "mozilla" => add_mozilla_roots(&mut root_cert_store),
       "system" => {
         let roots = load_native_certs().expect("could not load platform certs");
         for root in roots {
@@ -53,4 +55,45 @@ pub fn get_root_cert_store_provider(
   Ok(Arc::new(ValueRootCertStoreProvider::new(
     root_cert_store.clone(),
   )))
+}
+
+#[cfg(test)]
+mod tests {
+  use deno::deno_tls::rustls::pki_types::Der;
+  use deno::deno_tls::rustls::pki_types::TrustAnchor;
+
+  use super::*;
+
+  fn add_test_root(
+    root_cert_store: &mut RootCertStore,
+    subject: &'static [u8],
+  ) {
+    root_cert_store.roots.push(TrustAnchor {
+      subject: Der::from_slice(subject),
+      subject_public_key_info: Der::from_slice(b"test-key"),
+      name_constraints: None,
+    });
+  }
+
+  fn has_root_with_subject(
+    root_cert_store: &RootCertStore,
+    subject: &[u8],
+  ) -> bool {
+    root_cert_store
+      .roots
+      .iter()
+      .any(|root| root.subject.as_ref() == subject)
+  }
+
+  #[test]
+  fn add_mozilla_roots_appends_to_existing_store() {
+    let mut root_cert_store = RootCertStore::empty();
+    add_test_root(&mut root_cert_store, b"system");
+    let root_count_before = root_cert_store.roots.len();
+
+    add_mozilla_roots(&mut root_cert_store);
+
+    assert!(root_cert_store.roots.len() > root_count_before);
+    assert!(has_root_with_subject(&root_cert_store, b"system"));
+  }
 }


crates/deno_facade/cert_provider.rs
```diff
@@ -1,42 +1,46 @@
...
+      .expect("Failed to add platform cert to root cert store");
+  }
+}
+
+fn get_root_cert_store_with(
+  ca_stores: &[String],
+  mut add_mozilla_roots: impl FnMut(&mut RootCertStore),
+  mut add_system_roots: impl FnMut(&mut RootCertStore),
+) -> Result<RootCertStore, AnyError> {
   let mut root_cert_store = RootCertStore::empty();
```
kallebysantos: Couldn't it be simpler?
Any special reason why receiving it as closures?

I think the match expression below could just call the original fn directly

crates/deno_facade/cert_provider.rs
```diff
@@ -50,7 +54,103 @@ pub fn get_root_cert_store_provider(
...
+
+  fn root_cert_store_for(ca_stores: &[&str]) -> RootCertStore {
+    let ca_stores = ca_stores
+      .iter()
+      .map(|store| store.to_string())
+      .collect::<Vec<_>>();
+    get_root_cert_store_with(
+      &ca_stores,
+      |root_cert_store| add_test_root(root_cert_store, b"mozilla"),
+      |root_cert_store| add_test_root(root_cert_store, b"system"),
```
kallebysantos: Following my comment above, I couldn't get the idea about passing it down as closures using custom test functions.

AyaanFaisal21:
Good point. I used the closures so the original regression test would be deterministic, since system certs can vary across machines, and the closures made it easy to inject predefined test roots. However, I see that weaving test-only injection into production adds unnecessary bloat/abstraction.

I simplified the production code so the match arm now calls `add_mozilla_roots` directly, while the existing system cert-loading path remains inline. I also adjusted the regression test to stay separate and focus on the root cause: `add_mozilla_roots` appending to an existing `RootCertStore` instead of replacing it, so previously-added system roots are still preserved.

Verified with:

- `cargo test -p deno_facade add_mozilla_roots_appends_to_existing_store`
- `cargo test -p deno_facade`

AyaanFaisal21:
I noticed the workspace `cargo test` job failed in `-p base --test integration_tests` with several secure/TLS cases reporting `certificate has expired`. The `deno_facade` tests, including `cert_provider::tests::add_mozilla_roots_appends_to_existing_store`, passed in the same run.

This looks separate from the root-store append change, but I’m willing to investigate if you think it is related or want it addressed in this PR.

kallebysantos:
Hey mate,
Yes, it seems a problem not related to your PR, as it's failing on my local(main) branch as well.

> but I’m willing to investigate

Thank you, any help is welcome 💚

kallebysantos:
Seems we got this sorted already, https://github.com/supabase/edge-runtime/pull/707

kallebysantos:
Hey @nyannyacha, do you have any idea why theses tests are failing?
```
req_failure_case_op_cancel_from_server_due_to_cpu_resource_limit
req_failure_case_op_cancel_from_server_due_to_cpu_resource_limit_2
test_tmp_fs_usage
```
It doesn't seem related to the PR changes

reaganmcf:
Is there a timeline on this getting merged? We ran into this ourselves recently, and are doing the fix you mentioned in our ENV for the time being.

PR #722, opened 2026-08-27
fix: reclaim isolate host RSS on oneshot user-worker retirement

### LINKED ISSUE #719: Retired user workers are not fully reclaimed — RSS grows ~2.5 MiB per retirement with activeUserWorkersCount: 0 (v1.74.3, policy = oneshot)
### Summary

Under `policy = "oneshot"` the runtime retires one user worker per request, but host memory from retired isolates is only partially reclaimed. RSS grows roughly linearly with **retirement count** while the main worker's JS heap stays flat, so the container ends up holding hundreds of megabytes with **nothing running**.

This is not JS-heap growth and it is not application state — `/_internal/metric` reports `activeUserWorkersCount: 0` and a ~10 MiB JS heap while the container sits at 755 MiB RSS.

### Environment

- `public.ecr.aws/supabase/edge-runtime:v1.74.3` (current release at time of writing)
- Started by the Supabase CLI as part of a local stack, Docker Desktop on macOS, 7.75 GiB VM
- `policy = "oneshot"` (the same growth occurs under `per_worker`, roughly 2× faster)
- Functions are ordinary TypeScript handlers using `Deno.serve` and `npm:` specifiers

### Measurement

Container restarted for a clean baseline, warmed with one request, then driven with a realistic workload (an application's integration test suite). Readings are `docker stats` RSS alongside `/_internal/metric`:

|              | RSS       | `retiredUserWorkersCount` | `activeUserWorkersCount` | `receivedRequestsCount` | JS heap used |
| ------------ | --------- | ------------------------- | ------------------------ | ----------------------- | ------------ |
| before       | 265.6 MiB | 101                       | 0                        | 112                     | 8.38 MiB     |
| after        | 755.8 MiB | 300                       | 0                        | 312                     | 9.93 MiB     |
| **delta**    | **+490.2 MiB** | **+199**             | 0                        | +200                    | **+1.55 MiB** |

**≈ 2.46 MiB retained per retired user worker**, against a JS heap that moved 1.55 MiB in total.

Container forensics immediately after: `RestartCount=0`, `OOMKilled=false`, `Status=running` — this is steady-state growth, not a crash-and-restart artifact.

The memory is reclaimed only by restarting the container: a `docker restart` drops it straight back to the ~25–85 MiB baseline with no other change.

### Reproduction note that may save you time

**A trivial request loop does not reproduce this well.** Driving 100 requests at an endpoint that returns early (an unauthenticated `401`) produced a large one-off jump for the initial module-graph compile and then essentially plateaued:

```
0 req →  83 MiB   |  25 →  255 MiB  |  50 →  259.4 MiB  |  75 →  259.9 MiB  |  100 →  265.8 MiB
```

That is ~160 KiB/request in steady state — easy to dismiss as noise. The ~2.46 MiB/retirement figure only appears under requests that actually exercise the module graph and make outbound calls. Any repro harness needs realistic handlers, not a loop against a cheap endpoint.

### Why it matters in practice

Because growth tracks request count rather than time or concurrency, a long-lived container degrades steadily. We have separately observed it stop serving **below** its cgroup limit — at ~2.6 GiB of a 3 GiB cap, with `RestartCount=0`, `OOMKilled=false`, `Status=running`, every invocation failing `InvalidWorkerCreation: worker did not respond in time`. Because it never reaches the cap the kernel never kills it, so a `unless-stopped` policy never fires and the container stays up in a non-serving state indefinitely. That failure mode is harder to handle than a crash, since request retries all fail identically.

### What we ruled out

- **Not the JS heap** — `mainWorkerHeapStats.usedHeapSize` stays ~8–10 MiB throughout.
- **Not live workers** — `activeUserWorkersCount` is `0` at every reading above.
- **Not application state** — a container restart reclaims everything with no code or config change.
- **Not policy-specific** — `oneshot` roughly halves the rate versus `per_worker` but does not stop accumulation; memory carries across runs either way.

### Current workaround

Cap the container and recycle it on a memory threshold before it reaches the non-serving state. We restart above 1.5 GiB and then warm it with one request before releasing traffic, since a cold runtime's first invocation can otherwise exceed client timeouts.

### Happy to help

We can share `/_internal/metric` traces at additional points on the curve, test a patched image, or run a more targeted repro if you can suggest one — please say what would be most useful.

### Possibly related

#717 also comes from a self-hosted deployment and also concerns user-worker lifecycle, though from the `per_request` side (worker caps and secret refresh) rather than reclamation of already-retired isolates. Mentioning it in case the two share an owner.

samvallad33:
Oneshot retirement already drives `activeUserWorkersCount` to 0. Host RSS still grows ~2.5 MiB per isolate.

The worker is gone from the pool; the isolate's host pages are not. V8 leaves committed heap on the pinned worker thread, and that thread's allocator never returns those pages (`mi_collect` / `malloc_trim` / `malloc_zone_pressure_relief`). `EdgeRuntime.miCollect()` only purges the main worker.

This decommits the isolate on termination and drops host allocator pages in `DenoRuntime::drop` after `JsRuntime` dispose, on the same thread that owned the isolate.

`test_oneshot_retirement_drops_user_worker` checks the pool count hits 0 after oneshot shutdown. It does not measure RSS.

Fixes #719

### DIFF
--- a/crates/base/src/runtime/mod.rs
+++ b/crates/base/src/runtime/mod.rs
@@ -403,7 +403,8 @@ pub struct DenoRuntime<RuntimeContext = DefaultRuntimeContext> {
 
 impl<RuntimeContext> Drop for DenoRuntime<RuntimeContext> {
   fn drop(&mut self) {
-    if self.conf.is_user_worker() {
+    let is_user_worker = self.conf.is_user_worker();
+    if is_user_worker {
       self.js_runtime.v8_isolate().remove_gc_prologue_callback(
         mem_check_gc_prologue_callback_fn,
         Arc::as_ptr(&self.mem_check) as *mut _,
@@ -412,11 +413,20 @@ impl<RuntimeContext> Drop for DenoRuntime<RuntimeContext> {
 
     cleanup_js_runtime(&mut self.js_runtime);
 
+    if is_user_worker {
+      self.js_runtime.v8_isolate().low_memory_notification();
+    }
+
     unsafe {
       ManuallyDrop::drop(&mut self.js_runtime);
     }
 
     self.drop_token.cancel();
+
+    // Drop isolate host pages on the worker thread; pool counts already hit 0.
+    if is_user_worker {
+      ext_runtime::reclaim_host_memory();
+    }
   }
 }
 
--- a/crates/base/src/utils/test_utils.rs
+++ b/crates/base/src/utils/test_utils.rs
@@ -14,6 +14,7 @@ use anyhow::Context;
 use anyhow::Error;
 use either::Either::Right;
 use ext_event_worker::events::WorkerEventWithMetadata;
+use ext_runtime::SharedMetricSource;
 use ext_workers::context::MainWorkerRuntimeOpts;
 use ext_workers::context::Timing;
 use ext_workers::context::UserWorkerRuntimeOpts;
@@ -217,7 +218,7 @@ impl TestBedBuilder {
   }
 
   pub async fn build(self) -> TestBed {
-    let ((_, worker_pool_tx), pool_termination_token) = {
+    let ((metric_src, worker_pool_tx), pool_termination_token) = {
       let token = TerminationToken::new();
       (
         worker::create_user_worker_pool(
@@ -269,6 +270,7 @@ impl TestBedBuilder {
       pool_termination_token,
       main_termination_token,
       main_worker_surface,
+      metric_src,
     }
   }
 }
@@ -278,6 +280,7 @@ pub struct TestBed {
   pool_termination_token: TerminationToken,
   main_termination_token: TerminationToken,
   main_worker_surface: worker::WorkerSurface,
+  pub metric_src: SharedMetricSource,
 }
 
 impl TestBed {
--- a/crates/base/src/worker/supervisor/v8_handler.rs
+++ b/crates/base/src/worker/supervisor/v8_handler.rs
@@ -50,6 +50,7 @@ pub extern "C" fn v8_handle_termination(
   }
 
   if data.should_terminate {
+    isolate.low_memory_notification();
     isolate.terminate_execution();
   }
 }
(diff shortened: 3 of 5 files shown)


da1suk8:
I built both `main` (9fcc152) and this branch and measured RSS across oneshot retirements, following the methodology from #719 — a realistic handler that exercises the module graph (`npm:@supabase/supabase-js` + `npm:zod`) and makes an outbound `fetch` (kept hermetic by targeting the runtime's own health endpoint), driven sequentially. As the issue notes, a trivial handler does not reproduce the growth.

**Environment**: linux/arm64 (Docker Desktop on macOS, 64 GiB VM), container capped at 3 GiB, `start --main-service ./examples/main --policy oneshot`, `memoryLimitMb: 150` per worker (the example main's default). RSS = `VmRSS` of PID 1 in the container; worker counts from `/_internal/metric`.

Results after 300 retirements (sampled every 25; `activeUserWorkersCount` was 0 and the main-worker JS heap stayed flat at ~11 MiB at every sample, matching #719):

|  | RSS @ warm | RSS @ 300 retirements | growth | per retirement |
|---|---|---|---|---|
| main @ 9fcc152 | 189.6 MiB | 1292.9 MiB | +1103.3 MiB | ~3.7 MiB |
| this PR | 129.4 MiB | 301.1 MiB | +171.7 MiB | ~0.6 MiB |

Tail slope (retirements 150→300): ~837 KiB/retirement on main vs ~201 KiB/retirement with this PR. All 600 requests across both runs returned 200.

Latency cost of the reclaim: 50 sequential requests post-warm-up averaged 93 ms/req (main) vs 97 ms/req (this PR) — within run-to-run noise for a single pass in my setup.

Two observations:

- Growth per retirement scales with workload weight: with this heavier module graph, main retains ~3.7 MiB/retirement versus the ~2.46 MiB reported in the issue — consistent with its note that cheap handlers don't reproduce it.
- Some residual growth remains with this PR (~0.6 MiB/retirement overall, ~0.2 MiB at the tail). It may be allocator high-water behavior; `per_worker` policy (which the issue reports grows ~2× faster) isn't covered by my run or by the PR's test, so it may be worth checking as a follow-up.

Harness (repro function + measurement script) and raw CSVs: https://gist.github.com/da1suk8/53947f7182ab4aa9f4c5799ef4e67bd6 — happy to run other configurations.

PR #724, opened 2026-08-27
feat(fs): support Deno.copyFile for S3-backed file systems

da1suk8:
## Summary

Adds `Deno.copyFile` and `Deno.copyFileSync` support for S3-backed file systems.

- Copies objects with S3's `CopyObject` API without downloading data locally
- Maps missing source objects/buckets to `NotFound`
- URL-encodes copy-source object keys correctly
- Adds unit and integration test coverage for async and sync copies

## Limitation

This implementation relies on S3 `CopyObject`, which supports source objects up to 5 GiB.
Copying larger objects will require multipart copy support in a follow-up change.

## Validation

- `cargo fmt --all -- --check`
- `cargo metadata --no-deps --format-version=1 --locked --offline`
- `cargo test -p fs --lib copy_`

### DIFF
--- a/crates/fs/impl/s3_fs.rs
+++ b/crates/fs/impl/s3_fs.rs
@@ -819,7 +819,7 @@ impl deno_fs::FileSystem for S3Fs {
     self
       .client
       .copy_object()
-      .copy_source(copy_source(&source_bucket, &source_key))
+      .copy_source(to_encoded_copy_source_value(&source_bucket, &source_key))
       .bucket(destination_bucket)
       .key(destination_key)
       .send()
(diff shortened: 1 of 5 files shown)


crates/fs/impl/s3_fs.rs
```diff
@@ -2028,6 +2068,21 @@ fn try_get_bucket_name_and_key(path: PathBuf) -> FsResult<(String, String)> {
   ))
 }
 
+/// Builds the `x-amz-copy-source` value required by S3's CopyObject API.
+///
+/// The bucket/key separator and key path separators must remain intact, while
+/// each key segment is URL-encoded so names containing spaces or reserved
+/// characters resolve to the original S3 object.
+fn copy_source(bucket: &str, key: &str) -> String {
```
kallebysantos: I suggest renaming this method to something like `to_encoded_copy_source_value` as it may confuse with other things like `client.copy_source()`

da1suk8: Good point. I renamed the helper to `to_encoded_copy_source_value` to clarify that it returns the URL-encoded value passed to the SDK's copy_source method.

kallebysantos (approved):
Thanks for contributing 💚

PR #731, opened 2026-09-02
fix(runtime): allow Deno.copyFile in user workers

da1suk8:
## Summary

The `cargo test` job on main has been failing since #724 merged ([failing run](https://github.com/supabase/edge-runtime/actions/runs/33414290182/job/99561131906)): user workers expose the Deno fs API surface through an allowlist in `bootstrap.js`, and that change wired `copyFile`/`copyFileSync` into the S3 file system without adding them to the allowlist — so both still resolved to the blocklist stub, and the new integration test fails against any S3 backend with `PermissionDenied: Deno.copyFile is blocklisted`. The PR's own CI never caught this because fork PRs skip the S3 integration tests (the storage secrets are not available to them).

This allows `copyFile` unconditionally and `copyFileSync` only while the runtime is initializing, consistent with the other sync fs APIs (`writeFileSync` etc.); the `copyFileSync`-in-handler case in `test_copy_file` now asserts that rejection instead of expecting success. The san-check lists are updated accordingly.

It also makes the fs integration tests diagnosable: `assert_status()` surfaces the response body on a status mismatch and the fixtures return the error message in the body — a bare throw in a user worker turns into an opaque plain-text 500, and worker console output never reaches CI logs, which is why the original failure could not be diagnosed from the run alone.

## Validation

- Reproduced the exact CI failure and verified the fix against MinIO locally, then against a real Supabase Storage project: all 8 fs integration tests pass, including the 50 MiB upload-limit test.
- `test_user_worker_permission` passes with the updated san-check lists.
- Since fork PRs skip the S3 integration tests, I also ran the full CI on my fork with my own Supabase Storage credentials: [green run](https://github.com/da1suk8/edge-runtime/actions/runs/33580098635) — fmt, clippy, and the full test suite including `test_copy_file` against real Supabase Storage. (That run adds one fork-only CI commit on top of this branch: GitHub-hosted runners instead of Blacksmith, and org-token-gated steps skipped.)

### DIFF
--- a/crates/base/test_cases/user-worker-san-check/.blocklisted
+++ b/crates/base/test_cases/user-worker-san-check/.blocklisted
@@ -2,8 +2,6 @@ chmod
 chmodSync
 chown
 chownSync
-copyFile
-copyFileSync
 createSync
 fdatasync
 fdatasyncSync
--- a/crates/base/test_cases/user-worker-san-check/.whitelisted
+++ b/crates/base/test_cases/user-worker-san-check/.whitelisted
@@ -9,3 +9,4 @@ readTextFileSync
 readFileSync
 makeTempDirSync
 makeTempFileSync
+copyFileSync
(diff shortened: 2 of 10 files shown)


nyannyacha (approved):
LGTM

PR #732, opened 2026-09-02
feat(base): stamp project ref into the outbound user agent

satyarohith:
## What

Outbound requests made by a user worker carry the ref of the project it belongs to, in the `User-Agent`:

```
Deno/2.1.4 (variant; SupabaseEdgeRuntime/1.74.0; ref=abcdefghijklmnopqrst)
```

The ref comes from a new `projectRef` value on the worker create context:

```ts
EdgeRuntime.userWorkers.create({ servicePath, context: { projectRef: "abcdefghijklmnopqrst" } })
```

Workers created without it behave exactly as before.

## Why

There is no way to tell which project a request leaving a function came from, so reported abusive traffic cannot be attributed.

## How

- Requests that send no `User-Agent` get the runtime's, ref included.
- Requests that set their own keep it, with the `(variant; …; ref=…)` comment (valid per [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110#name-user-agent)) appended — a function cannot shed the ref. A header that already carries the comment is left alone, so redirect hops don't accumulate copies.
- Refs that don't fit safely in a header comment (anything outside `[A-Za-z0-9_-]{1,64}`) are dropped with a warning rather than escaped.

Covered paths: `fetch` (including `Deno.createHttpClient` and redirects), `node:http(s)`, `node:http2`, and websocket handshakes. Raw `Deno.connect` sockets have no HTTP layer and are not covered.

Note for consumers: a function can pre-fake a `ref=` comment in its own `User-Agent`. The real comment is appended after it, so always parse the last one.

### The one vendored change

`deno_fetch::Options::request_builder_hook` becomes `Option<Arc<dyn Fn(&mut HeaderMap) -> …>>`. A closure, because the ref differs per worker and a plain `fn` has no way to receive it (the hook gets no `OpState`). A header map, because `node:http2` builds `Request<()>` while fetch builds `Request<ReqBody>`, and a `dyn Fn` cannot be generic over the body type. The stamping logic itself lives in `crates/base/src/runtime/user_agent.rs`, so a vendor upgrade only re-applies the type.

## Tests

Unit tests for the append rules and ref validation, plus integration tests against a server that echoes back the `User-Agent` it received, covering `fetch`, `node:http`, `node:http2`, and the websocket handshake.

```
cargo test -p base --lib user_agent
cargo test -p deno versions::
cargo test -p base --test integration_tests user_agent
```

### DIFF
--- a/crates/base/src/runtime/mod.rs
+++ b/crates/base/src/runtime/mod.rs
@@ -111,6 +111,7 @@ use tokio_util::sync::CancellationToken;
 use tracing::debug;
 use tracing::instrument;
 use tracing::trace;
+use tracing::warn;
 use tracing::Instrument;
 use tracing::Span;
 
@@ -743,7 +744,16 @@ where
         let project_ref = context
           .get("projectRef")
           .and_then(serde_json::Value::as_str)
-          .and_then(deno::versions::sanitize_project_ref);
+          .and_then(|it| {
+            let sanitized = deno::versions::sanitize_project_ref(it);
+            if sanitized.is_none() {
+              warn!(
+                project_ref = it,
+                "invalid project ref is not stamped into the user agent"
+              );
+            }
+            sanitized
+          });
 
         let extensions = vec![
           deno_telemetry::deno_telemetry::init_ops(),
--- a/crates/base/src/runtime/mod.rs
+++ b/crates/base/src/runtime/mod.rs
@@ -758,8 +758,7 @@ where
           deno_canvas::deno_canvas::init_ops(),
           deno_fetch::deno_fetch::init_ops::<PermissionsContainer>(
             deno_fetch::Options {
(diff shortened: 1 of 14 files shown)


satyarohith:
The failing test looks unrelated:
https://github.com/supabase/edge-runtime/actions/runs/33611631917/job/100187903474?pr=732#step:16:1336
```
---- test_copy_file stdout ----
thread 'test_copy_file' panicked at crates/fs/tests/integration_tests.rs:184:7:
assertion `left == right` failed
  left: 500
 right: 200
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    test_copy_file
```

nyannyacha:
yeah seems related to https://github.com/supabase/edge-runtime/pull/731

nyannyacha (approved):
LGTM

PR #734, opened 2026-09-05
fix(fetch): preserve explicit trace context during propagation

### LINKED ISSUE #739: `traceparent` header rewritten on Edge Function-to-Edge Function invoke
# Bug report

> **TL;DR:** `traceparent` sent via `supabase.functions.invoke()` from one Edge Function to another is replaced by a platform-generated value before the callee sees it, breaking cross-function trace correlation. External inbound requests preserve it fine.

- [x] I confirm I have searched the
  - [x] Docs: https://docs.supabase.com
  - [x] Discussions: https://github.com/supabase/supabase/discussions
  - [ ] Discord: https://discord.supabase.com

## Describe the bug
When calling an Edge Function from another via `supabase.functions.invoke(...)` the `traceparent` header set on the outgoing
request is replaced by a different value before the target function receives it.
The receiving function sees a completely new `traceparent` with a different `trace-id` and `parent-id`. This is breaking trace correlation between the two Edge Functions.

This does not happen on inbound requests from external sources. This appears to be specific to function-to-function invokes inside the Supabase edge platform.

The replacement `trace-id` values appear to be time based, because we saw sequential trace-ids, e.g. `01a067d4d2d97b8aa324d2d094fd7d12`.

## To Reproduce
1. Deploy two Edge Functions, e.g. `caller` and `callee`.
2. In `caller`, invoke `callee` with an explicit `traceparent` header:
   ```ts
   const traceparent = "00-b1e669305668dc82c96cc31ce2e6cd99-1a5b74eca03f769f-01";
   await supabase.functions.invoke("callee", {
     headers: { traceparent, tracestate: "" },
     body: JSON.stringify({ hello: "world" }),
   });
   ```
3. In callee, log the incoming traceparent:
  Observe: the value logged in callee is not the value sent by caller. The trace-id is different (e.g. `01a067d4d2d97b8aa324d2d094fd7d12` and `01a06b8e23f57237952000924f945fda`).

## Expected behavior
The `traceparent` header sent by the caller should reach the callee unchanged, as it does on the inbound path from external clients. 
This is implied by [Client-side tracing](https://supabase.com/docs/guides/observability/client-side-tracing)
> the resulting trace_id flows through Supabase services and appears in API Gateway and Edge Function logs

## Screenshots
none

## System information
- OS: Supabase Edge
- Supabase-JS: 2.112.3
- Supabase CLI: 2.116.0
- Deno CLI: 2.9.6

## Additional context
- tracePropagation is not enabled on the client — the header is set explicitly in the headers option of functions.invoke().
- Verified no ContextManager is pre-registered in the user worker: `context.setGlobalContextManager(new syncLocalStorageContextManager().enable())` returns `true`.
- Verified no TracerProvider is registered: a probe span created via `trace.getTracer("probe").startSpan("probe")` has an all-zero trace id.
- `OTEL_DENO=true` is not active in the user worker.
- Custom headers in invoke survive unchanged.

## Workaround
Send the trace context in a custom-named header alongside the standard one. The custom header is not rewritten. On the receiver side, prefer the custom header over `traceparent` when present:
```ts
await supabase.functions.invoke("callee", {
  headers: {
    traceparent,
    tracestate,
    "x-app-traceparent": traceparent,
    "x-app-tracestate": tracestate,
  },
  body,
});

mandar1045:
## Tl;dr
Fixes https://github.com/supabase/edge-runtime/issues/739

Preserves caller-supplied `traceparent` and `tracestate` headers during function-to-function invocations.

## What changed
- In `vendor/deno_fetch/26_fetch.js`, skip injecting the trace context propagator if the request already contains an explicit `traceparent` header.
- Keeps `traceparent` and `tracestate` paired together from the caller, preventing the runtime from appending a second trace context or attaching mismatched runtime `tracestate`.
- Automatic trace propagation continues to work when no explicit header is provided.

## Validation
- `cargo test -p base --test integration_tests test_fetch_preserves_explicit_trace_context` (passed)
- Added 3 regression tests checking exact traceparent matching, comma prevention, and tracestate pairing.

### DIFF
--- a/crates/base/test_cases/trace-context-callee/index.ts
+++ b/crates/base/test_cases/trace-context-callee/index.ts
@@ -1,0 +1,8 @@
+Deno.serve((req: Request) =>
+  Response.json({
+    traceparent: req.headers.get("traceparent"),
+    tracestate: req.headers.get("tracestate"),
+    baggage: req.headers.get("baggage"),
+    custom: req.headers.get("x-app-traceparent"),
+  })
+);
--- a/crates/base/test_cases/trace-context-caller/index.ts
+++ b/crates/base/test_cases/trace-context-caller/index.ts
@@ -1,0 +1,23 @@
+const traceparent = "00-b1e669305668dc82c96cc31ce2e6cd99-1a5b74eca03f769f-01";
+
+Deno.serve(async (req: Request) => {
+  const origin = new URL(req.url).origin;
+
+  const headers = {
+    traceparent: traceparent,
+    tracestate: "foo=bar",
+    "x-app-traceparent": traceparent,
+  };
+
+  const response = await fetch(`${origin}/callee`, {
+    headers,
+    body: JSON.stringify({ hello: "world" }),
+    method: "POST",
+  });
+
+  const data = await response.json();
+  return Response.json({
+    sent: traceparent,
+    received: data,
+  });
+});
--- a/crates/base/test_cases/trace-context-main/index.ts
+++ b/crates/base/test_cases/trace-context-main/index.ts
@@ -1,0 +1,19 @@
+Deno.serve(async (req: Request) => {
+  const name = new URL(req.url).pathname.split("/").pop();
+  if (name !== "caller" && name !== "callee") {
+    return new Response("Not found", { status: 404 });
+  }
+
+  const worker = await EdgeRuntime.userWorkers.create({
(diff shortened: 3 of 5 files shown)


mandar1045:
Addresses #739 with integration tests included. Would appreciate a review whenever someone gets a chance

